PR #1, opened 2015-04-01
real fft, AB3, renaming and organizing

crocha700:
I've made two main changes: 

* Implemented functions that call fft, which can be numpy fft or fftw – this might be useful if we want to implement/test other stuff in the future (Because of overhead communications time, numpy is more efficient if nx is small; for nx >256 fftw is faster). I also change fft with real fft that explores the symmetry of the fourier coeffs given that the physical variables are real.  The nx=64 example is 30% faster.

* Changed the time-step to a third-order AB. In the two-layer model there's no problem of storage, so that   it's convenient to use AB3 because it's more accurate. 

* Added KE and eddy turn over time print out. I implemented two functions to compute these diagnostics.  I could use the get_diagnostic etc, but there's a normalization problem we should address first. To get correct units and a true spectrum (in the sense that the area under the surface is the total energy) we have to normalize the fourier coeffs. (see test_fft2.py).  I prefer to do that only when computing the spectrum in order not to degrade the accuracy. This has to be done for almost all diagnostics if we want to have correct units.

* I also added a notebook briefly describing notation and a couple of tests.

My short term todo list:

* Two  new classes: 1) equivalent barotropic model and 2) Eady (aka Blumen) model. I propose implementing these in separate classes. With these, we'll have the three simplest QG models. A future general could must recover results from these models as special cases. I already have code for this – I just want to make a smooth transition.

* Save snapshots of the solution into a file (likely netcdf). We probably only need q1 and q2 as we can always recover the other fields from the pv.

* Start a sphinx  documentation tree


rabernat:
Cesar, this is excellent! I have been tied up with teaching for the past week, but I am really looking forward to having a look at what you have done here!

rabernat:
By the way, I think the right approach towards making the model more general is to define a general class in which you specify the number of layers and the inversion operator, and then to subclass it for specific cases (e.g. 2-layer, Eady, barotropic, multi-layer QG, etc.)

crocha700:
Yeah, this might be a better approach; I'll try to find time to work on that over the weekend.  This may help for the what I'm thinking for the long-term, which is to do the integration in fortran. The Python class would set up all parameters (including the inversion matrix), initial conditions, etc. And then pass these on to fortran to do the heavy integration work.

PR #10, opened 2015-05-10
Bt model

crocha700:
BT model subclass. Still need to add some diagnostics, but it seems to be working pretty fine.


rabernat:
Nice work Cesar! This is awesome.

PR #24, opened 2015-05-15
Add sqg model

francispoulin:
Hello,

Today I spent a bit of time writing my first piece of code for pyqg. I basically took the bt model and modified it ever so slightly to solve the surface QG model. Note that I assume that the domain is semi-infinite, following Held et al. (1996), but it would be easy to modify it to have a bounded domain. That is something we will want to do but at least this is a step in the right direction (albeit of distance epsilon).

A few things to point out.
1) There is certainly room for improvement so please, if you happen to look at it, don't be shy with your criticisms.

2) I added a buoyancy parameter, Nb, but I have it set to one for the moment.

3) This code is only marginally different than bt_model. We might want to consider if there is a clean way of merging these into one. I suspect that later on we will want to have an option for an SQG layer above one or two QG layers. Do we really want different files for each or can we be clever and have parameters that we set which determine what we include.

4) I used the default colormap from the McWilliams. It looks great for the turbulence case but I don't think it is ideal since most of the vortex is white.

(this is my second attempt at doing this, maybe this will work)


crocha700:
Still trying to merge into 'master'...

rabernat:
Don't merge anything into master please. The master branch is only for functional, tested "releases" of the code. Let's just keep everything in develop.

PR #41, opened 2015-09-01
cleaned Model docstring

rabernat:


rabernat:
Also removed some unused kwargs in Model.__init__

Issue #49, opened 2015-09-02
Provide sqg example


rabernat:
We need an example for sqg.

rabernat:
Closed by #67.

PR #67, opened 2015-09-04
wrote sqg example and did minor changes

francispoulin:
I did some minor corrections here and there in what Ryan wrote.  

Then I wrote an example for the sqg case.  I realized that the parameters were not the same as in the paper and now they do reproduce the correct results.  Also, they pictures are much cooler!  I am tempted to change the final time to 40, what they showed in the paper since it's pretty nice to look at.  But that will slow things down so I haven't done so.

Please check it out and let me know what you think.  If you wanted to ignore my minor changes no problem.

A couple of things that I'd like to comment on.

First, to build the docs you need to type "make html".  Ryan said it was "build html", which was maybe a typo?
Second, I had to use jupyter-nbconvert since that's all I could find on anaconda, which I installed using conda.  Not sure if we wanted to mention that or not?


examples/test_sqg.py
```diff
@@ -5,9 +5,9 @@
 
 # the model object
 year = 1.
-m = sqg_model.SQGModel(L=2.*pi,nx=128, tmax = 100*year,
-        beta = 0., Nb = 1., H = 1., rek = 0., rd = None, dt = 0.01,
-                     taveint=year, use_fftw=True, ntd=8)
+m = sqg_model.SQGModel(L=2.*pi,nx=512, tmax = 40.005,
+        beta = 0., Nb = 1., H = 1., rek = 0., rd = None, dt = 0.005,
+                     taveint=1, ntd=8)
```
rabernat: I don't think 8 threads is appropriate for an example. We should probably not assume that users have a multicore machine.

rabernat:
@francispoulin, thanks so much for this contributions.

There are a couple of issues that have to be resolved before this can be merged.

- All of the documentation should be written in [restructured text](http://sphinx-doc.org/rest.html) format (.rst). The files in your pull request are all .html. You can translate your ipython notebooks into .rst by running ``nbconvert --to rst notebook.ipynb``.

- You should remove all the .ipynb_checkpoints/* files from the pull request. These are temporary files stored by ipython and do not belong in the repository.

To implement these changes, use ``git rm`` and ``git add`` as necessary, and then make a new commit. Then refresh your pull request by running

```bash
git push --force origin revised_sqg
```

examples/test_sqg.py
francispoulin: Ok, 8 is probably over the top.  My laptop only has 4 so I don't know where I got this from.  That being said, your barotropic example uses 4 and so using 4 above seems consistent.  
Now that I'm using 512x512, when I use 4 threads, it uses almost 3 of them, which is pretty good I think and does speed things up.

docs/examples/two-layer.rst
```diff
@@ -3,7 +3,7 @@ Two Layer QG Model Example
 ==========================
 
 Here is a quick overview of how to use the two-layer model. See the
-:py:class:`pyqg.QGModel` api documentation for further details.
+:py:class:``pyqg.QGModel`` api documentation for further details.
```
rabernat: This points to a larger issue in the documentation workflow. In sphinx, ```  py:class:`pyqg.QGModel` ``` has a special meaning which links to the relevant part of the API documentation. But if you write this in a notebook and use ipynb, the quotes get escaped to ```  py:class:``pyqg.QGModel`` ``, which breaks the links. The way I fixed this last time was to manually edit the .rst file, but that is not a long-term solution.

rabernat:
Ok, the rst files look good. There is still one html file though (docs/equations/notation_twolayer_model.html). Could you remove that?

Also, could you rename examples/test_sqg.py to something other than test_*? The prefix test_ has a special meaning for the test suite.

rabernat:
Perfect. Closes #49.

PR #70, opened 2015-09-06
modified sqg example and added equations

francispoulin:
I didn't get the same output that Ryan had on the sqg example.  I modified it anyhow and hope it looks better now.

I also added the equations part. I copied and pasted what I had in the example.  I don't suppose people have a problem with that?


docs/examples/sqg.rst
```diff
@@ -34,7 +34,7 @@ the buoyancy, :math:`b`, is hydrostatic balance
 .. math::
 
 
-   b = - \partial_z \psi. 
+   b = \partial_z \psi. 
 
  Using the fact that the Potential Vorticity is exactly zero in the
```
rabernat: The problem is that line 39 is indented. So it is interpreted as still part of the math block. I am surprised you didn't get this problem. Did you run ``make html`` to test your documentation changes?

rabernat:
Great, thanks for this quick update!

A few issues:

- The bug in docs/examples/sqg.rst is still there. (See my note above.)
- When I built the docs, I got a warning ``/Users/rpa/RND/Public/pyqg/docs/equations/notation_sqg_model.rst:: WARNING: document isn't included in any toctree``. This made me realize that you did not add an entry to equations.rst pointing to your new document. So please add that.
- You need to remove .ipynb_checkpoints/notation_twolayer_model-checkpoint.ipynb. (Are you adding your files with ``git add *``? Better to add them individually in the future to avoid this recurring problem.)

rabernat:
Nice! Looks perfect.

PR #73, opened 2015-09-08
Sqg equations

crocha700:
An update in the sqg equations doc.

### DIFF
--- a/docs/equations/notation_sqg_model.ipynb
+++ b/docs/equations/notation_sqg_model.ipynb
@@ -18,13 +18,13 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "The evolutions equations is\n",
+    "The evolutions equation is\n",
     "\n",
     "$$\n",
     "\\partial_t b + J(\\psi, b) = 0\\,,  \\qquad \\text{at} \\qquad z = 0\\,,\n",
     "$$\n",
     "\n",
-    "where $b = \\psi_z$ is buoyancy.\n",
+    "where $b = \\psi_z$ is the buoyancy.\n",
     "\n",
     "The interior potential vorticity is zero. Hence\n",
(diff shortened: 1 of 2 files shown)


francispoulin:
For some reason I still can't get a nice view. 
One thing I did notice is that you have
"The evolutions equations is\n",
Maybe remove the two "s" and have "The evolution equation is\n"?

rabernat:
@francispoulin are you on board with Cesar's changes here?

rabernat:
Just for the sake of discussion, what do you think about putting the model equations right in the docstring, as I did with the two-layer model?
http://pyqg.readthedocs.org/en/stable/api.html#pyqg.QGModel

If we go this route, we might be able to eliminate the "equations solved" section of the docs. We could have one detailed page about the numerical approach (inversion, pseudo-spectral advection, timestepping etc.) and then leave the individual model details in the docstrings.

francispoulin:
From what I can tell by looking at the source code I am happy with the changes.  Thanks @crocha700 for improving this.  But I can't see the finished product so I'm guessing a little bit.  If someone accepts these changes then I can view them afterwards and double check.

rabernat:
Ok, let's keep this for now and consider refactoring the documentation as a future possibility.

francispoulin:
Ok, I figured out that Firefox has problems with the equations but chrome does it beautifully.

@crocha700 , now I can see your equations.  My comments are below:

-) As I said before, "The evolutions equations is:" should be "The evolution equation is"
-) In our model N is constant but it does not have to be.  Joe LaCasce has done some work with non-constant N.  So have I but I haven't published anything, unlike Joe.
-) Before the 2nd z=0 I think you should have "at" not "and"

Looks great!

crocha700:
Those typos were fixed in b098c37 and  bbcf9aa.

PR #78, opened 2015-10-08
Layered model, topography, etc

crocha700:
Guys,

I'm merging my multi_layer branch into develop. I'm still working on documentation and testing, but I decided to merge so that you could test it and provide some feedback. New features include:

* N-Layer model (layered_model)
* Arbitrary topography
* A function to perform baroclinic linear stability analysis (this assumes flat bottom and constant U; it is easy to implement a constant bottom slope).
* A function to calculate the vertical modes and deformation radii (for every baroclinic model we now have a ``stretching matrix'').
* Logger 

qg_model.py still works but the idea is that this will replace it with N=2 as a special case.

The notebook  docs/examples/layered.ipynb showcases some of these features in a 3-layer system. Note that to implement topography I had to make changes in the kernel, so  maybe you need to rebuild the kernel after you checkout develop (and rebuild again once you go back to master...).


rabernat:
Wow, this is a lot of stuff! Fantastic!

There is too much to try to review this PR online, so I will just merge it (just into develop of course).

My initial impression is that more tests are probably needed, given the number of new features that have been added.

Issue #86, opened 2015-10-16
Bug in the kernel: missing negative sign


crocha700:
Working on the diagnostics for the layered model, I got confused by the negative sign we use in  http://pyqg.readthedocs.org/en/stable/examples/two-layer.html#plot-diagnostics

I believe this is a bug in the kernel. A missing negative sign in the calculation of the tendencies: 

```{python}
                for i in range(self.Nk):
                    # overwrite the tendency, since the forcing gets called after
                    self.dqhdt[k,j,i] = ( self._ik[i] * self.uqh[k,j,i] +
                                    self._il[j] * self.vqh[k,j,i] +
                                    self._ikQy[k,i] * self.ph[k,j,i] )
```

This should be negative since we are solving

```math_def
q_t = - J(psi,q) 
```

Note that the bottom drag has the correct sign.

rabernat:
A sign error in this core routine would be a pretty big bug. Are you sure? @MFJansen can you weigh in on this?

crocha700:
I think this was introduced in the kernel. It wasn't there in previous versions:
https://github.com/pyqg/pyqg/blob/b5db4bc8d3590c120a5cf5aee3d9bfc24c45903c/pyqg/model.py

See lines 305-308

rabernat:
Hmm... If so, it did not break the tests. Despite our problems with machine-precision reproducibility, I was careful about the tests when switching to the new kernel. I made sure that the solution on my specific machine did not change when switching to the cython kernel.

So there must be some kind of symmetry thing going on here that makes this problem cancel out.

rabernat:
Relevant lines.

Original:
https://github.com/pyqg/pyqg/blob/b5db4bc8d3590c120a5cf5aee3d9bfc24c45903c/pyqg/model.py#L308

Current kernel:
https://github.com/pyqg/pyqg/blob/master/pyqg/kernel.pyx#L373

crocha700:
Yes, this seems only to change the sign of the field! But when we compute diagnostics, say Re[psi^* J(psi,q)], then it shows up!

crocha700:
Where do we stand on this? I fixed this bug in the branch #87 I'm merging into develop.

I also added some documentation regarding the diagnostics: see docs/equations/notation_layered.ipynb --- that also describes N=2 as a special case for the diagnostics. If you focus on the first term on the RHS, you'll see that is the exact same term we compute in qg_model.py. No need to take the negative sign before plotting.

[...]

PR #87, opened 2015-10-20
Multi layer: diagnostics, documentations, and testing.

crocha700:
This also adds documentation regarding the multi-layer model and its diagnostics, adds a test using the multi-layer model and the same parameters as the two-layer model test, and   **fixes the bug** described in #86


rabernat:
@crocha700  - Can you make the output for the logging less verbose so that the test output is easier to read?

pyqg/test_bt.py
```diff
@@ -0,0 +1,47 @@
+import datetime as dt
+import numpy as np
+import matplotlib.pyplot as plt
```
rabernat: We should not import from matplotlib in any of the tests.

rabernat:
@crocha700 - it's great you are putting so much work into this! I have to admit that these huge commits are very hard to review. There are at least 5 completely separate features added here, and your last one was the same. Maybe for the future, consider splitting things up into multiple smaller PRs. (I recognize that this might be difficult when all the features are interdependent.)

pyqg/kernel.pyx
```diff
@@ -295,7 +295,7 @@ cdef class PseudoSpectralKernel:
         self.ifft_qh_to_q()
         # input might have been destroyed, have to re-assign
         self.qh[:] = b_view
-        
+
```
rabernat: The fact that you are mixing whitespace cleanup with major changes to the core computational routines makes the code quite hard to review.

crocha700: Yes, this is really annoying. Sorry. Will do a much bette job next time.

crocha700:
Ok, that test_bt was an old thing that I added accidentally in this commit. Should pass everything now. Also logger looks better now.

crocha700:
OK, I'm merging this into develop. This was my last changes in the layered-model, which should be working OK. Please test it out if you have any experiments in mind.

rabernat:
No! please don't merge without updating whats-new.rst in the docs.

crocha700:
@rabernat does the documentation builds automatically in readthedocs?

rabernat:
Too late again...

crocha700:
Sorry. OK, next time I'll assign somebody to review/merge to avoid this confusion again...

rabernat:
The documentation is here:
http://pyqg.readthedocs.org/en/develop/examples.html

You can check the build status here:
https://readthedocs.org/projects/pyqg/builds/

rabernat:
@crocha700, I'm very happy you are working so hard on this. The last thing I want to do is slow you down or discourage you.

My main request is that you keep your PRs more "atomic", i.e. one feature change per PR, in the future. This will make them easier to review. You might also want to consider [rebasing your PRs](https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request) when they contain tons of individual commits.

crocha700:
Fair enough. I'm actually rebasing my other branch "Eady model", which implements the stacked Eady model with the classic Eady model as a special case. I promise that that PR will be smoother.

crocha700:
@rabernat I just learned that we can easily revert merged PRs (there's a button called revert just next to the merged message). Would it be useful to revert this?

rabernat:
Nope, just create a new PR with the whats-new updates. I think it is important to keep track of the changes we are making in the docs.

-R

PR #91, opened 2015-10-22
Updates whatsnew

crocha700:
This simply updates "whats-new" with the main new features that were recently merged into develop.

### DIFF
--- a/docs/whats-new.rst
+++ b/docs/whats-new.rst
@@ -4,6 +4,12 @@ What's New
 v0.1.4 (? ? 2015)
 --------------------
 
+Added topography.
+
+Added automated linear stability analysis.
+
+Added multi layered model subclass. 
+
 Fixed bug in _calc_diagnostics. Now diagnostics start being averaged at
 tavestart.
 
--- a/docs/whats-new.rst
+++ b/docs/whats-new.rst
@@ -4,7 +4,7 @@ What's New
 v0.1.4 (? ? 2015)
 --------------------
 
-Fixed `bug <https://github.com/pyqg/pyqg/issues/86>`__ related to the sign of advection terms.
+Fixed bug related to the sign of advection terms (:issue:`86`). 
 
 Added new diagnostics. Those include time-averages of u, v, vq, and the spectral divergence of enstrophy flux.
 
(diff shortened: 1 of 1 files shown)


docs/whats-new.rst
```diff
@@ -4,6 +4,12 @@ What's New
 v0.1.4 (? ? 2015)
 --------------------
 
+Added topography.
+
+Added automated linear stability analysis.
+
+Added multi layered model subclass. 
+
```
rabernat: This doesn't seem complete. What about?
* the bug fix related to the sign of advection terms
* the new diagnostics
* the changes to sqg model
* the new logger

Anything else I'm missing?

crocha700: What changes in the SQG model?

rabernat: Maybe I'm confused about the sqg model...I thought you changed something there too. Maybe I was thinking of this: f48d4e8f2a8dee86a434c2a6612e66a27271d38b

The point is, I don't know! All I know is that your first list was incomplete.

crocha700: Topography does not work in that semi-infinite model, but I had to make little changes just to be compatible since the topographic term is computes in the kernel.

crocha700:
How about now? I also figured it would be useful to have links to the issue tracker, particularly when it refers to bug fixing.

docs/whats-new.rst
```diff
@@ -4,7 +4,19 @@ What's New
 v0.1.4 (? ? 2015)
 --------------------
 
-Fixed bug in _calc_diagnostics. Now diagnostics start being averaged at
+Fixed `bug <https://github.com/pyqg/pyqg/issues/86>`__ related to the sign of advection terms.
```
rabernat: We have [extlinks](http://sphinx-doc.org/latest/ext/extlinks.html#confval-extlinks) enabled in sphinx, so you should be able to just do ```:issue:`86` ```

crocha700: Good call.

rabernat:
Nice!

So is this ready to be a new release? How stable are the new features?

crocha700:
There are some little bugs in the documentation regarding the rendering of latex equations (I'm on it). The features seem stable to me. That been said, I was the only person that tested it out!

rabernat:
So then this needs to be separated into 0.1.4 and 0.2.

crocha700:
Sure. Let's merge that first, then I'll update this PR.

rabernat:
And I forgot to merge this into the 0.1.4 release... :weary:
I guess I need sleep.

I will merge it, but it is too late to actually include it in the release, which has already been published to pypi.

rabernat:
Actually, it should have been merged into master, not develop. double fail.

Issue #85, opened 2015-10-16
Add stochastic forcing


crocha700:
This would be useful for problems in which we want to force the system without imposing a background velocity.

francispoulin:
Agreed! I have been in contact with Geoff and I was hoping to try and reproduce some of his plots from his textbook on forced one-layer QG turbulence.  It shouldn't be too difficult and would be a lot of fun.  Unfortunately, I'm teaching 2 courses this term but will try and do what i can.

rabernat:
:+1: But maybe this is beyond 0.2?

crocha700:
Wallis & Maltrud was exactly the paper I had in mind @francispoulis.

Fine with me to add latter. Do we have a due date for 0.2?

PR #93, opened 2015-10-22
Fixes bug in advection

crocha700:
Fixes the sign error in advection calculation #85 .


crocha700:
@rabernat This merging from the upstream "master" simply resolves that conflict that we had when both of us changed .travis.yml

rabernat:
Are we worried about the sign of the advection in the diagnostics / examples? This is what originally led you to the bug, no?

crocha700:
Fixed.

rabernat:
The spectral energy transfer figure looks quite different. Do we understand why? Does this point to some (untested) differences when the sign of advection is changed?

rabernat:
Or is that due to the ``tavestart`` bug?

crocha700:
This differences likely stem from the ```tavestart`` bug`. The new estimates are much smoother.

crocha700:
Or not. Here's a [quick test](http://git.io/vWCZu) 

How robust are averages? In other word, are these differences statistically significant?

rabernat:
> How robust are averages? In other word, are these differences statistically significant?

It would be great if our spectra came with error bars.

rabernat:
Explain how I should interpret your test?

crocha700:
I set ```tavestart = 0``` in that example.

rabernat:
Well I am at a loss here. Our tests are incapable of discriminating between these different cases. Do you think we should just merge and hope for the best? Or should we try to understand the differences better?

crocha700:
We should attempt to understand if that's a real difference. The [original example](http://git.io/vWCh5) (Summer 2014) seems closer to diagnostics fixed above.

 While changing the sign does not change the total potential enstrophy, that bug may ensue other differences. If that change in sign was the only difference, then we should expect no difference in the diagnostics in a statistical sense.

crocha700:
>> How robust are averages? In other word, are these differences statistically significant?

>It would be great if our spectra came with error bars.

So, the error bar are small. [Here](http://git.io/vWCZu) I calculated the error bars assuming the number of degrees of freedom is large, and therefore assuming that the square of the Fourier coefficients are normally-distributed. The error bars are really small.

crocha700:
QG modeler generally do not present error bars in these estimate simply because we average enough to render those errors insignificant. So I'd say we incorporate the ```spectral_error``` as a method in ```diagnostic_tools````

rabernat:
> So I'd say we incorporate the spectral_error as a method in diagnostic_tools`
:+1

As for the problem at hand, what do you think we should do? I am inclined to get the opinion of @mfjansen.

crocha700:
I really don't know if we expect diagnostics to be exactly the same after fixing that bug. They are quantitatively similar, though. Let's wait for @MFJansen to weigh in. 

We definitely need more unit tests. We'll be glad as the project grows.

MFJansen:
Can somebody clarify to me which two results exactly we are comparing?

rabernat:
@MFJansen Go to the "files changed" tab of the PR and scroll down to the file docs/examples/two-layer_files/two-layer_12_0.png. You will see an A/B comparison of the spectral flux diagnostics before and after this bug fix.

MFJansen:
Thanks @rabernat!

These don't look terribly different to me, but maybe still a bit more different than I would have expected.
I think the safest way to estimate the statistical uncertainty is to run an ensemble - just a few simulations should be enough to give us an idea. The initial conditions are pseudo-random by default, but I'm pretty sure the seat for the random number generator will always be the same right now (at least for a given machine), so we would need to change that to have a pseudo-random seed (based on system time or something) and then just run a few runs with different noise realizations as initial conditions, and see how different the spectral energy budgets look.
     
The more I think about it the less I think the sign error would cause a difference here (beyond the "statistical error"). If I see it correctly, there is a symmetry in the equations in that the advection term is the only quadratic, and all other terms are linear. So if we change the sign of the advection term, I think )-q) should evolve like q would evolve otherwise. Since the initial condition here is just a little bit of noise changing the sign of that small noise shouldn't impact the statistical evolution.

rabernat:
> there is a symmetry in the equations in that the advection term is the only quadratic, and all other terms are linear.

I agree. I think it is effectively the same as reversing the direction of one of the spatial dimensions. Github markdown doesn't support math, but something like

J(psi, q) = dpsi/dx dq/dy - dpsi/dy dq/dx

If you make the transformation x -> -x, you get

-dpsi/dx dq/dy + dpsi/dy dq/dx

i.e. the sign is just reversed.

I am personally fine with merging this as is, without introducing new test, in order to get a bug fix out asap. (Did you see the email from Geoff? Are you on the [pyqg-dev mailing list](https://groups.google.com/forum/#!forum/pyqg-dev)?) In version 0.2, we can finally confront our inadequate testing regime. But I will defer to you on this decision.

crocha700:
Really like @MFJansen's idea. [Here are some calculations](http://git.io/vWlC8) with 10 different seeds. There is some spreading, in the spectral divergence of the fluxes, particularly at low wavenumbers. This spreading is larger than the difference between two simulations with the same seed but different ```tavestart```.

francispoulin:
Very nice work guys.  I'm impressed with how close things look with only 10 different seeds.  

One question, if we want to compare the statistics, would it be better to have a coarser grid and more members of the ensemble?  That might be easier to show convergence across different systems.

MFJansen:
@rabernat: Yes, that's actually a better way to think about it - particularly because the linear advection terms (that is background flow advection and beta term) were also affected.
Re the email from Geoff: I don't know how to do this best, but I think we should definitely make sure that this bug gets fixed in whatever is the latest "official" release ASAP.

@crocha700: Thanks for setting that up! This suggests to me that the differences we were seeing were with in the statistical noise.

rabernat:
> I think we should definitely make sure that this bug gets fixed in whatever is the latest "official" release ASAP.

If we are all comfortable with this fix, I can merge it right now and make a new release.

crocha700:
Fine with me.

MFJansen:
Yes, I think we should get this fix in.

In the long run we still need to work on implementing better tests though. If I'm seeing this right, we had Rossby waves traveling in the wrong direction - that probably shouldn't go unnoticed...

PR #98, opened 2015-10-23
changed license to MIT

### LINKED ISSUE #76: change license
We picked GPL somewhat randomly. Apparently it is not the best choice for a science application:
https://twitter.com/wesmckinn/status/651009426879594496

Some good advice here:
http://www.astrobetter.com/blog/2014/03/10/the-whys-and-hows-of-licensing-scientific-code/

rabernat:
closes #76


crocha700:
Wow, this is so much better. Direct to the point.

rabernat:
Ok @crocha700 you merge it then.

From now on, let's try to avoid merging our own PRs when possible.

crocha700:
Setup needs to be changed accordingly. Also, do we want to keep track of this in whats-new?

crocha700:
Travis build failed again. Is the time step too large in that test?

rabernat:
test_twolayer_qy.py and test_twolayer_qg_fftw.py are almost identical! Why do we have these redundant tests?

crocha700:
I don't know. Somehow I can't traceback the history of "test_twolayer_qg_fftw" on github.

MFJansen:
I believe test_twolayer_qg_fftw was introduced when we were hoping to get bitwise identical results when using similar fft routines. You can see that the "correct" value differs here, which I believe is because they were computed using fftw versus something else (I believe MKL). If my memory serves me correctly there used to be also a flag to turn on fftw, which I guess has disappeared when moving to cython. 
So in summary, given that we have given up on having bitwise reproducibility, and given that the two versions don't actually use different fft routines, test_twolayer_qg_fftw should be redundant.

rabernat:
> given that we have given up on having bitwise reproducibility

We have not given up! :stuck_out_tongue_winking_eye: We have just postponed.

> given that the two versions don't actually use different fft routines

I'm not sure if you guys even realize this, but I added back numpy fft support to the kernel. But the choice of fft routine needs to be made a compile time, not run time.

Anywy, I am going to remove the _fftw version.

rabernat:
I think this can be merged. We need to figure out why the travis-ci tests sometimes throw cfl errors, but that can be part of the larger testing refactor.

crocha700:
Yes, that's weird that it fails sometimes on travis. (It passes on my mac). 

I'll merge this.

crocha700:
I'm merging this. Tests don't pass in travis but do pass on my machines.

PR #101, opened 2015-10-25
use gcc instead of cc for openmp check

### LINKED ISSUE #100: check_for_openmp does not necessarily use the same compiler as cythonize
Everything works fine on linux server with CentOS. But a Mac OS El Capitan with gcc installed from  [Mac-hpc](http://hpc.sourceforge.net), ```check_for_openmp``` uses ```clang``` whereas ```cythonize``` appears to use ```gcc```.

If we simply comment out that test in setup.py, 

```python
extra_compile_args = []
extra_link_args = []

use_openmp = True
#if check_for_openmp() and use_openmp:
#    extra_compile_args.append('-fopenmp')
#    extra_link_args.append('-fopenmp')
#else:
#    warnings.warn('Could not link with openmp. Model will be slow.')    

extra_compile_args.append('-fopenmp')
extra_link_args.append('-fopenmp')
```

then everything works.

rabernat:
should fix #100.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,7 @@ from Cython.Build import cythonize
 import warnings
 import numpy as np
 import os
-import tempfile, subprocess, shutil   
+import tempfile, subprocess, shutil
 
 VERSION='0.1.4'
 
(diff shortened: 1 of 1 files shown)


crocha700:
Great. It works.

PR #103, opened 2015-10-26
Layered subclass

crocha700:
This adds a multi-layer subclass, with the two-layer model as a special case. I wrote the [equation part of the docs](http://git.io/vWwMz) in ```latex``` and converted to ```rst``` using [pandoc](http://pandoc.org):
```{bash}
crocha (layered *) equations $ pandoc -f latex notation_layered.tex -t rst >> notation_layered.rst
```
Standard diagnostics (spectra and divergence of spectral fluxes) are implemented. A 3-layer example is available [here](http://git.io/vWwMC).

PS: Note that the docs already feature topography, linear stability analysis, and standard vertical modes. Those implementations will follow in different PRs.


pyqg/layered_model.py
```diff
@@ -0,0 +1,322 @@
...
+    def _initialize_forcing(self):
+        pass
+        #"""Set up frictional filter."""
+        # this defines the spectral filter (following Arbic and Flierl, 2003)
+        # cphi=0.65*pi
+        # wvx=np.sqrt((self.k*self.dx)**2.+(self.l*self.dy)**2.)
+        # self.filtr = np.exp(-self.filterfac*(wvx-cphi)**4.)
+        # self.filtr[wvx<=cphi] = 1.
+
+    def _calc_diagnostics(self):
```
rabernat: It seems like this method should be in model.py, not the subclass.

pyqg/layered_model.py
```diff
@@ -0,0 +1,322 @@
...
+        self.Hi = np.array(H)
+        self.rhoi = np.array(rho)
+
+        super(LayeredModel, self).__init__(**kwargs)
+
+
+    ### PRIVATE METHODS - not meant to be called by user ###
+
+
+    def _initialize_stretching_matrix(self):
```
rabernat: Can you add units tests for this method?

pyqg/layered_model.py
```diff
@@ -0,0 +1,322 @@
...
+                    self.S[i,i]   = -self.f2/self.Hi[i]/self.gpi[i-1]
+                    self.S[i,i-1] =  self.f2/self.Hi[i]/self.gpi[i-1]
+
+                else:
+                    self.S[i,i-1] = self.f2/self.Hi[i]/self.gpi[i-1]
+                    self.S[i,i]   = -(self.f2/self.Hi[i]/self.gpi[i] +
+                                        self.f2/self.Hi[i]/self.gpi[i-1])
+                    self.S[i,i+1] = self.f2/self.Hi[i]/self.gpi[i]
+
+    def _initialize_background(self):
```
rabernat: And this one.

pyqg/layered_model.py
```diff
@@ -0,0 +1,322 @@
...
+        # the meridional PV gradients in each layer
+        self.Qy = self.beta - np.dot(self.S,self.Ubg)
+        self.Qx = np.dot(self.S,self.Vbg)
+
+
+        # complex versions, multiplied by k, speeds up computations to precompute
+        self.ikQy = self.Qy[:,np.newaxis,np.newaxis]*1j*self.k
+        self.ilQx = self.Qx[:,np.newaxis,np.newaxis]*1j*self.l
+
+    def _initialize_inversion_matrix(self):
```
rabernat: And this one.

rabernat:
The figure docs/examples/layered_files/layered_12_0.png looks kind of funky. Totally dominated by the lowest wavenumber.

rabernat:
@crocha700 overall this looks good. Much easier to see what you have done.

We just need tests for the inversion matrix stuff.

crocha700:
Thanks @rabernat. I'll work on those tests. Could you just clarify which method should be in model.py?

As for the spectral energy budget figure in the example, the dynamics there is similar to that in [Larichev & Held (1995)](http://www.gfdl.noaa.gov/bibliography/related_files/vdl9501.pdf), the energy cascade up to the domain scale (see their figure 4). We can try to come up with a different example given that this is going to the docs.

rabernat:
> Could you just clarify which method should be in model.py?

``_calc_diagnostics``

It doesn't actually do anything different from the same method in model.py:
https://github.com/pyqg/pyqg/blob/master/pyqg/model.py#L376

rabernat:
> As for the spectral energy budget figure in the example, the dynamics there is similar to that in Larichev & Held (1995): the energy cascades up to the domain scale (see their figure 4)

Yes, I just checked, and I agree that the figures are similar. So forget about that comment. But perhaps we should reference Larichev and Held explicitly?

crocha700:
I added the tests for the methods specific to the ```LayeredModel``` subclass. For the ```_init_background```, I also included assertions within the method. I also added the citation to the notebook, and made some tiny fixes here and there.

rabernat:
Your tests look good.

I should have mentioned this before, but I think it is preferable to use the [unittest framework](https://docs.python.org/2/library/unittest.html) for testing.

Would it be possible for you to wrap your tests following the example here?
https://github.com/pyqg/pyqg/blob/master/pyqg/tests/test_model.py

And if possible, split the different tests into separate methods?

I don't think this would be too hard. Then we can get this merged.

@MFJansen, @francispoulin, do you have any comments?

francispoulin:
I really like the LH95 test that @crocha700 put together.  I don't know the unittest framework at all so don't have much to say on that.  On Monday I plan to sit down and try and catch up.  When I see the tests first hand I might be able to give you more concrete advice.  

Also, on a very different note, a while ago I wrote something up to compute the linear stability of a one-layer QG model that has spatial variation.  I unfortunately (fortunately?) never included it in the core.  I know that @crocha700 has done a multi-layer version but I gather for horizontally uniform flows.  I would be interested to know where you think my test should go?  Maybe if I look at the new structure it might make more sense to me.

rabernat:
Beautiful work @crocha700!

PR #106, opened 2015-10-28
Update installation.rst for OpenMP on OSX

jamesp:
Updates to the Installation instructions:
- Added missing cython dependency to list at top of document
- clang gcc compiler on OSX Yosemite and El Capitan does not have openmp support (see discussion in issue #100).  Added a section with instructions on compiling with an alternative gcc compiler if openmp support is required.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -10,6 +10,7 @@ The only requirements are
 
 - Python 2.7. (Python 3 support is in the works) 
 - numpy_ (1.6 or later)
+- Cython (0.2 or later)
 
 Because pyqg is a pseudo-spectral code, it realies heavily on fast-Fourier
 transforms (FFTs), which are the main performance bottlneck. For this reason,
@@ -194,7 +195,12 @@ remember how.)
 
 Installing pyqg
 ---------------
-
+.. note::
+    The pyqg kernel is written in Cython and uses OpenMP to parallelise some operations for a performance boost.
+    If you are using Mac OSX Yosemite or later OpenMP support is not available out of the box.  While pyqg will
+    still run without OpenMP, it will not be as fast as it can be. See :ref:`advanced-install` below for more
+    information on installing on OSX with OpenMP support.
+    
 With pyfftw installed, you can now install pyqg. The easiest way is with pip:
 
 .. code-block:: bash
@@ -224,3 +230,52 @@ If you want to make changes in the code, set up the development mode:
 
 pyqg is a work in progress, and we really encourage users to contribute to its
 :doc:`/development`
+
+
+.. _advanced-install:
+
+Installing with OpenMP support on OSX
+-------------------------------------
+
+There are two options for installing on OSX with OpenMP support.  Both methods require using the Anaconda distribution of
+Python.
+
+1. Using Homebrew
+
+Install the GCC-5 compiler in ``/usr/local`` using Homebrew:
+
+.. code-block:: bash
+
+    $ brew install gcc --without-multilib --with-fortran
+
+Install Cython from the conda repository
+
+.. code-block:: bash
+
+    $ conda install cython
+
+Install pyqg using the homebrew ``gcc`` compiler
+
+.. code-block:: bash
+    
+    $ CC=/usr/local/bin/gcc-5 pip install pyqg
+
+
+2. Using the HPC precompiled gcc binaries.
+
+The `HPC for Mac OSX <http://hpc.sourceforge.net/>`__ sourceforge project has copies of the latest ``gcc`` precompiled for Mac OSX.  Download the latest version of gcc from the HPC site and follow the installation instructions.
+
+Install Cython from the conda repository
+
+.. code-block:: bash
+
+    $ conda install cython
+
+Install pyqg using the HPC ``gcc`` compiler
+
+.. code-block:: bash
+    
+    $ CC=/usr/local/bin/gcc pip install pyqg
+    
+
+    


crocha700:
@jamesp could you please provide a short and punchy description to this PR. I understand that's a simple update to the docs, but a little text would help folks that did not participate in the discussion #100  to understand what's being done here.  

An important update is the inclusion ``cython`` to the requirements list.

jamesp:
@crocha700 sure, no problem.  Now updated.

docs/installation.rst
```diff
@@ -224,3 +230,52 @@ If you want to make changes in the code, set up the development mode:
...
+
+
+.. _advanced-install:
+
+Installing with OpenMP support on OSX
+-------------------------------------
+
+There are two options for installing on OSX with OpenMP support.  Both methods require using the Anaconda distribution of
+Python.
+
```
rabernat: Why exactly do they require anaconda?

jamesp: On my iMac, the only way I have successfully managed to compile with openmp is with the anaconda distribution.  As mentioned in #100, when I use builtin python Cython tries to pass `-Wshorten-64-to-32` to the compiler as it is expecting `clang`, but this is not a valid flag for HPC gcc.

We can change the line requiring anaconda, it may be an issue just affecting me.

rabernat:
Ok well this seems reasonable.

MFJansen:
I'm confused. The docs still don't seem to mention anything about this. 
And honestly, I'm still unable to install the latest version of pyqg on my iMac.
I use El Capitan, xCode version 7.1 and a Canopy Python installation (so a pretty standard and up-to-date setup).
Here is what I get:
gcc -bundle -undefined dynamic_lookup -g -arch x86_64 -headerpad_max_install_names -arch x86_64 build/temp.macosx-10.6-x86_64-2.7/pyqg/kernel.o -o build/lib.macosx-10.6-x86_64-2.7/pyqg/kernel.so -fopenmp
ld: library not found for -lgomp
clang: error: linker command failed with exit code 1 (use -v to see invocation)
error: command 'gcc' failed with exit status 1

So, as discussed above, this does seem to have to do with openmp not being supported with clang. But I still don't see this or a workaround documented in the installation doc.

crocha700:
>I'm confused. The docs still don't seem to mention anything about this. 
 This is in the [latest version of the docs](http://pyqg.readthedocs.org/en/latest/installation.html). The stable version is related to the last release.

crocha700:
>gcc -bundle -undefined dynamic_lookup -g -arch x86_64 -headerpad_max_install_names -arch x86_64 build/temp.macosx-10.6-x86_64-2.7/pyqg/kernel.o -o build/lib.macosx-10.6-x86_64-2.7/pyqg/kernel.so -fopenmp
ld: library not found for -lgomp
clang: error: linker command failed with exit code 1 (use -v to see invocation)
error: command 'gcc' failed with exit status 1

What do you get when you do  ```$ which gcc``` ?

crocha700:
@jamesp Have you tried to install this with canopy?

MFJansen:
@crocha700 Thanks, I was just starting to realize that I just don't actually have the latest version. I'm confused by that too, but that has more to do with my limited understanding of github..

$ which gcc
/usr/bin/gcc

crocha700:
How did you install this ```gcc```? 

What if you do 
```
CC=/usr/bin/gcc python setup.py build_ext --inplace
```

MFJansen:
ok, I just realized that git was still pointing to the old repository under rabernat/pygq, so this is way not up to date. I should probably fix that first...

crocha700:
you can remove this link
```
$ git remote rm upstream
```
and then follow what's in the [docs](http://pyqg.readthedocs.org/en/stable/development.html#develpment-workflow).

MFJansen:
ok, this was the root of all my confusion. It seems to be working fine now.
Sorry about that, I guess that's what happens if you are out of the loop for too long...

crocha700:
Good. An important conclusion is that the approach described in the [docs](http://pyqg.readthedocs.org/en/latest/installation.html) works as good with canopy as with anaconda...so we need to change the docs accordingly.

MFJansen:
I actually only did the standard installation without openmp so far. I think the part in the docs that's specific to anaconda is for openmp support. I will see if I get that to work with Canopy and will report back.

MFJansen:
If I try to use gcc-5  from homebrew or the pre-compiled HPC libraries, I get the following error when installing pyqg:
In file included from /usr/local/lib/gcc/x86_64-apple-darwin15.0.0/5.2.0/include-fixed/syslimits.h:7:0,
                 from /usr/local/lib/gcc/x86_64-apple-darwin15.0.0/5.2.0/include-fixed/limits.h:34,
                 from /Applications/Canopy.app/appdata/canopy-1.5.5.3123.macosx-x86_64/Canopy.app/Contents/include/python2.7/Python.h:19,
                 from pyqg/kernel.c:24:
/usr/local/lib/gcc/x86_64-apple-darwin15.0.0/5.2.0/include-fixed/limits.h:168:61: fatal error: limits.h: No such file or directory
compilation terminated.
error: command '/usr/local/bin/gcc' failed with exit status 1

Any ideas?

rabernat:
@MFJansen This is a conflict between your canopy installation and your gcc version include directories. I don't understand the details of it. You are the only person using canopy as far as I know.

Can you build with your default system gcc (clang)?

Can you try with anaconda?

There is a potentially pretty large matrix of platforms / compilers / python environments to deal with here. I wonder how other cython projects deal with this.

MFJansen:
@rabernat Yes, I can built with the system default gcc, which is clang. But not with openmp.

I don't have Anaconda right now and have never used it before; but seems to be what the cool kids are using these days, so maybe I'll give that a try.

rabernat:
This seems to be pretty similar:
http://stackoverflow.com/questions/30624829/python-h-fatal-error-limits-h-no-such-file-or-directory-include-limits-h

rabernat:
Also this:
http://ask.sagemath.org/question/10672/did-anyone-else-face-a-problem-with-cython-after-upgrading-to-mavrick/
https://groups.google.com/forum/#!topic/sage-devel/5ne3SwEeYAE/discussion

crocha700:
I could compile ```pyqg``` with ```openmp``` using a canopy distribution of python – I had some difficulties with pyfftw though.

Our problem is akin to the problem of all compiled models. I think the best we can do is to have a more verbose doc with some "recipes" that worked for a specific system and python distribution, or on a specific computer, etc, as [the dedalus guys do](http://dedalus-project.readthedocs.org/en/latest/installation.html).

jamesp:
@crocha700 Writing a set of recipes of configurations we know to work sounds like a good approach.  It seems that it is probably only going to get harder in the future.  As Apple continue to make their software less open and more tailored to writing objective-c than c++ it's going to become harder and more confusing for new users to get involved.

Up until recently I had been a strong advocate for using builtin where possible with `pip` and `virtualenv` to build your own modules, but now I use `conda`.  I think it's great that Anaconda and Canopy are taking the pain out of the process of building software and making it much more accessible to users on all platforms.

Issue #71, opened 2015-09-06
instability of a QG jet


francispoulin:
Inspired by the recent advances in pyqg, I would like to put together another QG example.  Basically, I want to reproduce some of the results of Glenn's from 1987.

http://journals.ametsoc.org/doi/abs/10.1175/1520-0485(1987)017%3C1408:NWACVS%3E2.0.CO;2

He uses the reduced gravity BT QG model but one issue is that the streamfunction is not periodic in one direction.  One way to do this is to decompose the model into a basic state and perturbation

psi = Psi + psi'
q    = Q   + q'
u    = U   + u'

The fully nonlinear governing equations are then

partial_t q' + J(Psi, q') + J(psi', Q) + J(psi',q') = 0

and we can look for solutions that are doubly periodic.  

I know how to solve this in general but not sure how to fit this into the context of pyqg.  I looked in BT_model and the evolution equations are not there.

Ryan: any ideas how best to do this? 

Another option is to look how to use DCT in one direction.  I thought the first option would be easier but am very interested to know what you all think. 

If you think this is too much of a pain for the moment I don't have to do it now but I thought I'd mention it since it would be a nice example. Doing the linear stability would be nice and easy too.

rabernat:
This is a great suggestion. It seems like there are two different things being suggested here

1. Allowing for spatially variable background states
2. Enforcing different lateral boundary conditions

I think both should eventually be implemented. In particularly, 1 would allow us to include topography through the pv gradients, something I would be very interested in. 2 would open a whole new range of scenarios, including gyres and stuff.

It would be quite easy to implement 1. The model already adds a constant background U and Qy. The kernel would just have to be modified to allow these to be arrays instead of constants. An important part of this would be to keep things backwards compatible with the examples we already have. All of the action is in kernel.pyx.

As for 2, pyfftw simply does not support DCT. However, the fftw package itself does.
http://www.fftw.org/doc/Real-even_002fodd-DFTs-_0028cosine_002fsine-transforms_0029.html
In order to take advantage of this, we would need to bypass pyfftw and make our own cython wrapper for fftw. I'm sure it would be tricky, but there are plenty of examples to go on, e.g. [daedalus](https://bitbucket.org/dedalus-project/dedalus/src/150221b94a6a75712cb12098c2585287ffa8bffb/dedalus/libraries/fftw/fftw_wrappers.pyx?at=default). This would have the added benefit of removing the pyfftw dependency, which can be a pain for some people.

[...]

PR #108, opened 2015-10-30
Vertical modes

crocha700:
This implements simple methods to compute vertical modes and project solutions onto those modes. The [3-layer example](http://git.io/vlfrv) was updated to showcase the usage of these methods. Unit tests are implemented.


docs/examples/layered.rst
```diff
@@ -167,6 +167,56 @@ A snapshot and some diagnostics
...
+
+.. code:: python
+
+    m.vertical_modes()
+
+We can also project the solution onto the modes
+
+.. code:: python
+
+    pn = m.modal_projection(m.p)
```
rabernat: This is super cool! I like this way of doing it.

pyqg/model.py
```diff
@@ -205,7 +205,41 @@ def run(self):
...
+        self.radii = np.zeros(self.nz)
+        self.radii[0] = self.g*self.H/np.abs(self.f) # barotropic def. radius
+        self.radii[1:] = np.sqrt(1./evals[asort][1:])
+
+        # eigenstructure 
+        self.pmodes = evecs[:,asort]
+
+        # normalize to have unit L2-norm
+        Ai = (self.H / (self.Hi[:,np.newaxis]*(self.pmodes**2)).sum(axis=0))**0.5
+        self.pmodes = Ai[np.newaxis,:]*self.pmodes
```
rabernat: This code looks super clean. I love it.

rabernat: Could you add ``radii``  and ``pmodes`` to the class attributes docstring (line 60 of model.py)?

rabernat:
This is a great PR.

francispoulin:
@crocha700 this looks great!  A few very minor comments.

You define kd = 1/Ld to be the inverse wavenumber.  Technically, I think you should have a 2*pi in there.  Not to say that this distracts from the coolness of your results but I thought I would mention it.

Also, is there a paper you could cite?  I know that Glenn has an earlier paper for the 2-layer vs 2-mode that shows they are equivalent, more or less.  But the 3-layer/3-mode problem is richer and if we pointed the reader (maybe me ) to the paper that might help.

Finally, you compute the kinetic energy density in terms of layers.  If you have a modal decomposition, I presume  you could easily do that in terms of the vertical modes.  That might be interesting to see.

Great job!

crocha700:
Thanks for the reviews guys!

@francispoulin I  followed Glenn and many others, and defined the deformation radii as the square root of the inverse eigenvalue. I believe that that's a confusion that exists our field...To try to avoid this confusion, some (including Bill and I) call L_d = 2*pi/k_d the deformation length, as opposed to deformation radius  R_d = 1/k_d.  Just updated the docs with a citation to [Flierl 1978](http://www.sciencedirect.com/science/article/pii/0377026578900027).

As for the energy, I agree that you can define it modal wise energy, but since we're doing the advection in physical space (as opposed to modal space), there's no point in doing this online. The inclusion  of a modal subclass would be interesting, but that's a matter for a different  PR.

rabernat:
> The inclusion of a modal subclass

You mean one that actually solves the modal equations? Interesting idea.

rabernat:
> but since we're doing the advection in physical space (as opposed to modal space), there's no point in doing this online

I think that online diagnostics of modal KE would actually be very useful. This breakdown can't be recovered from the depth-space average, due to the coupling between modes. But that doesn't have to go in this PR.

crocha700:
Let me try to implement these diagnostics then (don't merge this yet). It is actually quite simple: the modal energy spectrum is simply

(wv2+kn^2)*|ph_n|^2, where ph_n is the projection of ph onto the nth mode.

crocha700:
>You mean one that actually solves the modal equations? Interesting idea.

This is the idea that Gleen introduced and [Hua and Hadivogel 1986](http://journals.ametsoc.org/doi/abs/10.1175/1520-0469(1986)043%3C2923%3ANSOTVS%3E2.0.CO%3B2) implemented. Andy and Ross also used a similar approaches for their theses (though they restricted themselves to the two-mode case). There are some advantages to the dynamics: Gleen argues that for "realistic" simulations one doesn't need to "calibrate" the modal mode (as opposed to matching deformation radii and interaction coefficients for layer models, etc.). [We've recently generalized this approach to correctly account for inhomogeneous  boundary conditions](http://git.io/vlOtO), but I somehow arbitrarily decided to pursue a highly accurate finite differences/Pade approximation in the vertical and never implemented those equations...Might be a good excuse to do it now.

crocha700:
OK, I included  modal KE and PE spectra to diagnostics, and updated the example.

francispoulin:
@crocha700 Thanks for clarifying about the radii vs length.  I see your point and will try and follow suit.

Also, thanks a lot for updating the example.  I'm glad it wasn't too much trouble and I do think that it's interesting.  It shows that at the large scales we have most of the energy is in the BT mode as we should expect.  Plotting the PE is also a nice touch.

Tomorrow I going to try and update my pyqg.  I haven't done it in a while and it's going through some big changes and I'm excited to see how things work.  I asked a question a while ago about including the linear stability analysis of the Bickley let.  I would like to include that and even do a multi-layer version.  True that we probably won't go higher than 4 layers or so but still, that would be interesting.

I also would like to do nonlinear simulations with pyqg.  Since the Bickley jet does not have periodic BCs, I think I would need to decompose the flow into a basic state and a perturbation and solve the nonlinear perturbation equations.  I know the equations, I'm just not sure how this will fit in the pyqg framework.

I realize this is not really part of issue 108 and maybe I should have started another one.  Sorry about that.

rabernat:
My only questions is whether we should be plotting isotropic spectra in the example.

crocha700:
OK, I updated the [example](http://git.io/vlfrv) to plot isotropic spectra. (I'll submit a different PR updating, testing and documenting those diagnostic_tools methods)

crocha700:
@francispoulin let's move the Bickley jet discussion to #71

rabernat:
Perfect!

Issue #83, opened 2015-10-12
need linear stability examples


rabernat:
We have this nice example notebook:
https://github.com/pyqg/pyqg/blob/develop/docs/examples/layered.ipynb

But it was not converted to rst
http://pyqg.readthedocs.org/en/develop/examples.html

PR #115, opened 2015-11-02
Baroclinic linear stability method

crocha700:
This implements a method to perform baroclinic linear stability analysis associated with a model class. A unit test that compares the numerical results with exact results is provided.  Equations and an [example](http://git.io/vlKOq) were added to the docs. Thus this addresses #83.

One thing that came up when implementing this method was how to best perform the [sorting to obtain the most unstable mode](http://git.io/vlKsa). I attempted to use ```argsort()``` .

```{python}
isort = evals.imag.argsort(axis=0)
evals = evals[isort]
self.omg = evals[-1]
```
but this implementation was ridiculously slow. Using ```argmax()``` was also a bit confusing and slow.  So I moved on to the this with a nested for loop, which turned out to be much faster. Another possibility is store all the modes and let the user play with it (this may be useful if one is interested in more than a single mode for every wavenumber). Suggestions welcome.


crocha700:
Just rebased master to include those small changes #111 and #113

pyqg/model.py
```diff
@@ -243,6 +243,40 @@ def modal_projection(self,p,forward=True):
...
+        evals, evecs = evals.T, evecs.T
+    
+        # sorting things this way proved way
+        #  more faster than using numpy's argsort() !
+        imax = evals.imag.argmax(axis=0)
+        for i in range(self.nl):
+            for j in range(self.nk):
+                self.omg[i,j] = evals[imax[i,j],i,j]
+                self.evec[:,i,j] = evecs[imax[i,j],:,i,j]
+
```
rabernat: Rather than adding ``omg`` and ``evec`` to the model as attributes, maybe it would be better to just return them from this method.

pyqg/model.py
```diff
@@ -243,6 +243,40 @@ def modal_projection(self,p,forward=True):
 
         return pt
 
+    def stability_analysis(self,bottom_friction=False):
+        """ Baroclinic linear instability analysis """
```
rabernat: A more verbose and correctly formatted docstring would be nice.

pyqg/model.py
```diff
@@ -243,6 +243,40 @@ def modal_projection(self,p,forward=True):
...
+
+        M = np.einsum('...ij,...jk->...ik',L4,(L3+Q).T)
+
+        evals,evecs = np.linalg.eig(M)
+
+        evals, evecs = evals.T, evecs.T
+    
+        # sorting things this way proved way
+        #  more faster than using numpy's argsort() !
+        imax = evals.imag.argmax(axis=0)
```
rabernat: Is this where you pick root with the fastest growth rate?

francispoulin:
@crocha700  I am surprised that a double look is faster than a build in function of python.  I don't know but I wonder if this is because you are sorting a 2D array. Is that right?  I wonder if it might be faster to sort long vectors.  I've used argmax before on big vectors and didn't notice any problem, but then again I never looked that closely.

pyqg/model.py
crocha700: Yes.  In this implementation I find the most unstable mode (max image part): ```axis=0``` corresponds to ```nz``` unstable modes.

crocha700:
@francispoulin I'm sorting a 3D array (```nz,nl,nk```) in a single direction.

francispoulin:
Ah, sorry.   I guess you found a reasonable solution so maybe it's better to stick with that.

rabernat:
This seems good to go too, no?

crocha700:
As far as I can tell, yes.

Issue #96, opened 2015-10-23
travis-ci master build is failing


rabernat:
How did we miss this? 
https://travis-ci.org/pyqg/pyqg/pyqg/builds/86957787
It looks like it failed on a cfl problem.

My tests are running fine, and the 0.1.4 release branch passed fine too. 
https://travis-ci.org/pyqg/pyqg/builds/86957835
(It should be identical to master at this point.)

What is going on?

rabernat:
Looks like it was a fluke. Just reran the build and it worked fine...
https://travis-ci.org/pyqg/pyqg/builds/86957787

Issue #62, opened 2015-09-03
Make tests faster


rabernat:
There are a lot of redundant tests. Plus they are slow. How long do we have to run to be confident the model is "right"?

crocha700:
The reference solutions, particularly ```bt``` and ```sqg```, are the killers:

```{python}
%%bash
for test in *.py
    do 
        stats="$(sh -c "time python $test -args &> /dev/null" 2>&1)"
        echo $test  "           "   $stats >> profile_test.out
    done
cat profile_test.out


test_advection.py             real 0m0.543s user 0m0.462s sys 0m0.067s
test_bt.py             real 0m23.476s user 0m23.215s sys 0m0.169s
test_fft2.py             real 0m0.194s user 0m0.123s sys 0m0.052s
test_fftw.py             real 0m11.442s user 0m11.329s sys 0m0.103s
test_import.py             real 0m0.022s user 0m0.012s sys 0m0.008s
test_layered.py             real 0m0.199s user 0m0.123s sys 0m0.052s
test_model.py             real 0m0.193s user 0m0.123s sys 0m0.051s
test_particles.py             real 0m0.197s user 0m0.126s sys 0m0.051s
test_sqg.py             real 0m21.505s user 0m21.332s sys 0m0.126s
test_twolayer_qg.py             real 0m4.884s user 0m4.764s sys 0m0.072s
test_vertical_modes.py             real 0m0.196s user 0m0.125s sys 0m0.051s
```

crocha700:
I'll submit a PR to change this.

Issue #90, opened 2015-10-21
Diagnostics: name/description and normalization


crocha700:
I know this may sound perfectionism for some. Current description of the diagnostics is misleading. These terms represent tendencies in the "spectral energy density". The spectrum has units of . The have units of (L^2 T^{-3}) per unit wavenumbe. For example, if the wanumber is in cycles per km, and energy in m^2/s^2, then this would be m^2 s^-3 / cpkm.

My main concern is:
1. The "flux terms"  are actually *divergence* of flux...To get the flux we must integrate in wavenumber.
2.  "APEgen" is actually represents the spectrum of the rate of potential energy generation.
3. "Total energy dissipation" should be the spectrum of mechanical energy dissipation through bottom drag.
etc.

Also, the implementation of the fft computes an [unnormalized](http://www.fftw.org/doc/The-1d-Discrete-Fourier-Transform-_0028DFT_0029.html#The-1d-Discrete-Fourier-Transform-_0028DFT_0029) transform. Note that because we're using pyfftw, this does not matter if we are just transforming back and forth. But,  to get the correct magnitudes for the diagnostics we need to normalize to normalize things by the size of the array... In particular for quadratic quantities, such as the energy spectral density, we need to normalize by  M^2, where M = Nx x Ny.

MFJansen:
I agree that the "flux terms" are really the (spectrally) local tendencies due to the nonlinear transfers - that is really spectral flux convergences. That should probably be clarified.

APEgen and EKEdiss I believe are actually be the total rates not the spectra though.

The normalizations may need to be double checked. I believe I had them worked out correctly at some point, but things may have changed since we are using different ffts.

PR #116, opened 2015-11-03
Faster tests

crocha700:
Currently the tests are too slow #62 

```{bash}
$ time nosetests 
(...)
----------------------------------------------------------------------
Ran 26 tests in 56.840s

OK

real	0m57.094s
user	0m56.450s
sys	0m0.391s
```

This PR makes some changes in the reference solutions "tests". First, it refactors the three reference solutions into a single file. Second, it changes the reference solutions for ```sqg``` and ```bt``` to make the tests run faster. Those were the bottle necks as per profile reported in #62. 

(I kept the ```two_layer``` reference solution the same since that was the original solution... But when running this tests many times on my computer the ```two_layer``` reference solution test fails every once in a while.This issue was previously reported in #96. I figure that the problem is that the time step is too large in that test.)

With these changes:
```{bash}
$ time nosetests 
(...)
----------------------------------------------------------------------
Ran 26 tests in 12.711s

OK

real	0m13.174s
user	0m12.684s
sys	0m0.279s
```

Thus these changes will make the suit of tests to run about 5 times faster.


pyqg/tests/test_twolayer_qg.py
```diff
@@ -1,112 +0,0 @@
...
-    
-    ## need to average these diagnostics
-    avg_diagnostic_results = {
-        'entspec': 5.703438193477885e-07,
-        'APEflux': 9.192940039964286e-05,
-        'KEflux': 1.702621259427053e-04,
-        'APEgenspec': 9.058591846403974e-05,
-        'KE1spec': 3.338261440237941e+03,
-        'KE2spec': 7.043282793801889e+01
-    }
```
rabernat: At some point we will want to put the diagnostics tests back in, no?

crocha700: We might, but with the currently values the inconsistency is wild:

```{python}
EKEdiss
   EKEdiss: 1.114800472490948e-10 
            6.368668363708053e-08 (desired)
APEgen
    APEgen: 1.260862629353291e-07 
            8.842056320175081e-08 (desired)
APEgenspec: 1.291480122136782e-04 
            9.058591846403974e-05 (desired)
   entspec: 8.172722867144368e-07 
            5.703438193477885e-07 (desired)
   APEflux: 1.347717697982667e-04 
            9.192940039964286e-05 (desired)
    KEflux: 2.653355719569081e-04 
            1.702621259427053e-04 (desired)
```

May be we could restart this test with a new reference solution.

MFJansen: Uhm, according to this EKEdiss has changed by over two orders of magnitude. I'm sure this is significant! The new values seem wrong to me, since EKEdiss is more than two orders of magnitude smaller than EKEgen. 
 
The other differences actually seem kind of significantly different to me as well. We should probably figure out what's going on. Unfortunately this test is a mess - remember issue #2? At the same time it's the only test that goes all the way back, so we should probably figure it out.
Reading through the discussion on issue #2 again, I realize that the conversation wondered quite a bit, and ended up being mostly about bitwise reproducibility. However, it looks like it started with inconsistencies that were clearly far greater.

rabernat: > it's the only test that goes all the way back

I just want to point out that, at this point, the current pyqg code has received far more scrutiny that the "original" python script. We have reproduced several classic papers and developed quite thorough tests of certain aspects of the code. There are still big gaps in the testing of course, especially regarding the diagnostics. But maintaining reproducibility of the "original" code might be a red herring at this point.

My first guess is that this order-of-magnitude issue is related to the normalization of the FFTs.

rabernat: Also there are similar order-of-magnitude differences between the "old values" (commented out in the code) and the current values. Does anyone remember when that got changed?

crocha700: >My first guess is that this order-of-magnitude issue is related to the normalization of the FFTs.

Agreed. The KE spectrum is now being normalized by ```m.M^2`` (the issue of normalization is not settled yet #90). This explains the KEdiss order-of-magnitude difference.

MFJansen: > I just want to point out that, at this point, the current pyqg code has received far more scrutiny than the "original" python script. We have reproduced several classic papers and developed quite thorough tests of certain aspects of the code.

I guess I'm a little less optimistic about this than you are. Let's not forget that until very recently we didn't realize that we had the sign of the advection term wrong (and in fact it was this experiment that made Cesar find this error - since he realized that the sigs of the diagnostics in the notebook had to be reversed.) 
Have we reproduced (quantitatively) classical two-layer results?
And as you are saying: I can't see any test of the diagnostics.  

Not to be a Debby Downer here, but I think we should take this seriously.

crocha700: >Have we reproduced (quantitatively) classical two-layer results?

I have reproduced Larchev & Held 1995 for a talk a gave in the Theory Seminar class that Paola is organizing this quarter. The simulation was done with the correct sign of advection. The diagnostics were computed offline because LH95 work with modal formulation.
[My_Talk_TheorySeminar_FA2015.pdf](https://github.com/pyqg/pyqg/files/26258/My_Talk_TheorySeminar_FA2015.pdf)

francispoulin: Nice talk @crocha700 !

Also, I think that the SQG run reproduces the results of Held et al., thanks to @crocha700  fixing something.

rabernat: > Not to be a Debby Downer here, but I think we should take this seriously.

Ok, yes, I agree. If it is not something simple like M^2, it will be quite a difficult task to track down the source of the disagreement. We will have to take the whole model apart and put it back together.

rabernat: If I roll back to e7c5ee2ed204c640400517e4c820f75918666185, the test passes.

rabernat: One issue is that the ``tavestart`` parameter is ignored in that early version.

```python
        # here is where we calculate diagnostics                                  
        if (self.t>=self.dt) and (self.tc%self.taveints==0):                      
            self._increment_diagnostics() 
```

If I set ``tavestart==0`` in the latest version, I get much better (but still not perfect) agreement
```
   EKEdiss: 7.502854837145852e-11 
            6.368668363708053e-08 (desired)
    APEgen: 8.571462649785877e-08 
            8.842056320175081e-08 (desired)
APEgenspec: 8.779580143694401e-05 
            9.058591846403974e-05 (desired)
   entspec: 5.519416835379740e-07 
            5.703438193477885e-07 (desired)
   APEflux: 8.986470103263862e-05 
            9.192940039964286e-05 (desired)
    KEflux: 1.768313116056767e-04 
            1.702621259427053e-04 (desired)
```

crocha700: Same here with ```tavestart=0 ```, but my results are slightly different than @rabernat's/

```   
    EKEdiss: 7.502521767193113e-11 
            6.368668363708053e-08 (desired)
    APEgen: 8.570574054449991e-08 
            8.842056320175081e-08 (desired)
APEgenspec: 8.778646755959426e-05 
            9.058591846403974e-05 (desired)
   entspec: 5.519110737187770e-07 
            5.703438193477885e-07 (desired)
   APEflux: 8.986199885636697e-05 
            9.192940039964286e-05 (desired)
    KEflux: 1.769198042761687e-04 
            1.702621259427053e-04 (desired)
```

rabernat: Regarding the magnitude of ``EKEdiss``:
```python
In [2]: 7.502521767193113e-11 * (32**2)
Out[2]: 7.682582289605747e-08
```
Which is now also in a similar ballpark.

crocha700:
and with canopy and numpy 1.9.2

```{python} 
  EKEdiss: 7.507421618592886e-11 
            6.368668363708053e-08 (desired)
    APEgen: 8.577384674904105e-08 
            8.842056320175081e-08 (desired)
APEgenspec: 8.784672251018768e-05 
            9.058591846403974e-05 (desired)
   entspec: 5.522155131808690e-07 
            5.703438193477885e-07 (desired)
   APEflux: 8.987687995232196e-05 
            9.192940039964286e-05 (desired)
    KEflux: 1.765612305511634e-04 
            1.702621259427053e-04 (desired)
```

pyqg/tests/test_twolayer_qg.py
MFJansen: @rabernat Good call on the tavestart! 
Since the model runs to right after the breaking of the unstable waves, it seems to me that the results are pretty sensitive to minor changes. It looks like machine-dependence probably can't explain the remaining differences, but a slight change in the numerics or the filter probably could.

francispoulin:
Looks like @landscape-bot is getting agreement between two numpy's to 3 or so significant digits.    Not great but at least within the right order of magnitude. 

If we wanted to verify whether we should still consider the "desired" value, I wonder if we could compare this with another model that someone has lying around?  If this model has been tested by others then we should be able to figure out the order of magnitude rather easily. 

Maybe this is too much work though.

pyqg/tests/test_twolayer_qg.py
MFJansen: Regarding the normalization. The new normalization of EKEdiss seems to be more right, as can be inferred by comparing to rek*EKE2, which should give the same answer save for an O(1) factor given by the fractional depth of the lower layer. The latter actually appears to be missing in code though. So I think there still seem to be some issues with the diagnostics and how layer depths are accounted for (originally the diagnostics assumed two layers of equal depth. I thought I generalized this at some point, but apparently I missed some things...) I will try to review the energy budget diagnostics sometime soon. There is at least an easy consistency check that can be made, which is to check if the budget closes - I thought this was the case for the spectral energy budget, but maybe not for the bulk energy budget...

crocha700:
I just restored the old content of ```test_two_layer_qg``` into ```test_reference_solns```. Thus, there is no change as far as that test is concerned. The goal of this PR is to simply organize this into a single tester and make the ```sqg``` and ```bt``` test faster.

We think should pick up this important discussion of ```test_two_ayer``` and other tests in different issues and PRs.

pyqg/tests/test_twolayer_qg.py
francispoulin: To make things more concrete, is there a section in Geoff's book that we can cite and refer to how this should be calculated for the layered equations?  That might be useful for us and for other users as well, who want to know what convention we are following.

rabernat:
> Looks like @landscape-bot is getting agreement between two numpy's to 3 or so significant digits.

What do you mean by this? lanscape doesn't run any tests. Do you mean travis-ci? If so, this test suite does not actually include the diagnostic checks we are debating.

pyqg/tests/test_twolayer_qg.py
MFJansen: I'm confused. From my reading of the code (whether right or not), these two should be identical:

In  [188]: m.get_diagnostic('EKEdiss')
Out[188]: 1.1141103759804538e-10

In  [189]: 2*m.rek*m.get_diagnostic('EKE')[-1]
Out[189]: 1.805314588712134e-10

But somehow they are not. Does anybody else see what I am missing?

rabernat: from model.py
```python
        self.add_diagnostic('EKEdiss',
            description='total energy dissipation by bottom drag',
            function= (lambda self:
                       (self.rek*self.wv2*
                        np.abs(self.ph[-1])**2./(self.M**2)).sum())
        ) 
```

MFJansen: @rabernat are you just providing some background information for others, or was that meant to be a response to my question? (If so I don't get it)

rabernat: It was in response to your question. That is how the ``EKEdiss`` diagnostic is calculated. EKE instead is
```python
        self.add_diagnostic('EKE',
            description='mean eddy kinetic energy',
            function= (lambda self: 0.5*(self.v**2 + self.u**2).mean(axis=-1).mean(axis=-1))
        )
```

Just trying to help debug...

MFJansen: Sorry, yes I saw that - I should have probably copied the relevant code segments in my comment in the first place. 
So why do these differ. It looks to me like they should be the same (safe for the factor 2*rek).

It looks like I can boil the issue down further. From just the last state of the model I get:

    (np.abs(m.p[-1])**2).mean()
    Out[221]: 1045198.6974801013
    
    (np.abs(m.ph[-1])**2./(m.M**2)).sum()
    Out[222]: 845712.10049875535

Shouldn't these be the same? (Parseval's identity)

crocha700: Gotta be careful with the Parseval's identity because we're using "real" FFT so we just have half of the coefficients. (That's why I created the "spec_var" method)

Here

```{python}
In [10]: (np.abs(m.p[-1])**2).mean()
Out[10]: 1038187.4151566233

In [11]: (np.abs(m.ph[-1])**2./(m.M**2)).sum()
Out[11]: 837904.39974625339
```
With ```spec_var```
```{python}
In [19]: m.spec_var(m.ph[-1])
Out[19]: 1038187.4151566236
```

Also, notice that the relative difference between our results is about 1%.

MFJansen: Thanks @crocha700 ! I must admit I don't quite see the factor of two in the difference, but your method, spec_var seems to do get it right.

However, that means that the EKEdiss diagnostic in the model is definitely broken! And probably this affects other diagnostics too.

crocha700: It is not only a factor of two...notice that the real fft for ```Nx=Ny``` even has the shape ```(Nx/2+1,Ny)``.

MFJansen: yeah, I got it (see my edit above)

MFJansen: I will go through the diagnostics and see / fix where this causes an obvious problem.

crocha700: Yes. I think much of this is related to the fact that we're now using real fft...

MFJansen: I need some help here: To compute the frictional dissipation correctly, I need to know the depth of the lowest layer - ideally as a fraction of the total depth, if we want to keep our energy budget in terms of volume average energies. Right now EKEdiss is a "core diagnostic", so it would work for all model types. If I see it correctly the way layer depths are handled differs between the two-layer and general layered model though, right? If I see it correctly I would have to use the vector self.Hi for the n-layer model, but this isn't defined for the two-layer model, right?

crocha700:
```{python}
In [6]: m.H2/m.H
Out[6]: 0.8
```

or

```{python}
In [9]: 1./(1+m.delta)
Out[9]: 0.8
```

MFJansen: Sorry, I guess I was unclear. I understand how to compute the layer depth ratio in the two layer model. 
The problem is that the EKEdiss diagnostic is implemented as a "core diagnostic", so the computation needs to work for all models (with the same code). I agree that it would be best to define Hi for the two layer model. It's not immediately obvious to me where this needs to be done (I'm not that familiar with the code anymore), but will have a look.

crocha700: yes I misunderstood you...so you can add something like

```
self.Hi = np.array([self.H1,self.H2])
```
just below [this](https://github.com/pyqg/pyqg/blob/master/pyqg/qg_model.py#L118)

MFJansen: Yeah, I actually just started editing it with the goal of getting rid of self.H1 and self.H2 completely. It's easy to do in qg_model.py; I'm just unsure if self.H1 and self.H2 are used somewhere else?

crocha700: apparently only [here](https://github.com/pyqg/pyqg/blob/master/pyqg/qg_model.py#L217) and [here](https://github.com/pyqg/pyqg/blob/master/pyqg/qg_model.py#L226)

MFJansen: Yes I found those. What I was unsure about is if they are used somewhere outside of qg_model.py. I don't see anything obvious...

crocha700: It shouldn't been used elsewhere outside this model subclass.

francispoulin: Sorry for being a bit lost in the discussions here.  It seems that the code that you are looking at is the legacy code that computes the two-layer diagnostics.  I know that @crocha700 computed an N-layer version. Does that have diagnostics as well? 

Has anyone tried computing the fluxes with the two approaches and the diagnostics for a given snapshot to see if they yield the same result?  Maybe the n-layer isn't quite that developed.  I haven't managed to figure that much out yet.

crocha700: The layered model does compute diagnostics as documented [here](http://pyqg.readthedocs.org/en/latest/equations/notation_layered.html#energy-spectrum).

MFJansen: The energy dissipation rate is a core diagnostic, so it's computed for any subclass model. 

The "APE generation rate" (really conversion from background flow to eddies) instead is so far only coded up as an additional diagnostic that exists in the two-layer sub-class.
@crocha700: you haven't coded up energy conversion rates for the n-layer model, right?

crocha700: [Here are the diagnostics](https://github.com/pyqg/pyqg/blob/master/pyqg/layered_model.py#L285) specific to the layered model subclass.

MFJansen: Nice! @francispoulin is right, we should be able to compare results from a strict 2-layer setup to results from the layered model with an identical two layer configuration. That might be a good sanity check.

MFJansen: I just pushed a minor fix, fixing the EKEdiss and APEgen diagnostics. 
I didn't do anything to the spectral diagnostics for now, since it seems less obvious what the "right" normalization here is.

MFJansen: Save for a factor of nx*ny, the results for APEgen and EKEdiss now appear to be very close (within less than 1%) of the values that were commented out as "old" in the test. Here's what I get:
     m.get_diagnostic('APEgen')*32**2
    Out[35]: 2.522085319118987e-07
    
    m.get_diagnostic('EKEdiss')*32**2
    Out[36]: 1.4800760866049312e-07

Notice that the summed up spectra are not expected to agree due to the fft-rfft switch. Maybe that's not the best thing to look at anyway. In particular not the integrated spectral flux convergences - which should just integrate to zero anyway, when done right.

francispoulin: That is a big step in the right direction!  Is that using the old 2-layer diagnostics or the new n-layer diagnostics?  Or maybe they are equal?

MFJansen: @francispoulin: EKEdiss is a "core diagnostic" (in model.py), so the same code is used for the two layer and n-layer model. APEgen on the other hand is specific to the two-layer subclass. However, since @crocha700 has already implemented the APE generation spectrum in the n-layer subclass, it should be easy to compute the total APE generation there as well (by "appropriately" summing over all wavenumber - which isn't quite as trivial anymore with the rfft format of the spectra, but it's not that hard either). One should also be able to move the computation of the APE generation into the "core", though that may require a little bit more work.

francispoulin: Thanks @MFJansen for spelling that out.  That helps.

MFJansen: ps: I just realized I broke the barotropic model (which doesn't have Hi) - I'm on it...

crocha700: @MFJansen next time please try to branch off of master, and then submit the fixes as a pull request (PR). This procedure is described in the [development workflow](http://pyqg.readthedocs.org/en/stable/development.html)  that @rabernat put together. This approach helps us keep track of what's changed, and avoid code breaking. Every PR is tests using [ci-travis](https://travis-ci.org). @rabernat put a lot of effort to make out editions and inclusion of new features as smooth as possible. Let's try to follow that workflow.

crocha700: @MFJansen Thanks for fixing the energy dissipation ad APEgen diagnostics. I think things look better now.  Here I get with the 2-layer subclass:

```{python}
In [43]: m.get_diagnostic('EKEdiss')*(32**2)
Out[43]: 1.4809381615971648e-07
```

MFJansen: oops, sorry, I'm not up to date with the procedures.

I guess differences \lessim 1% are to be expected for this test due to variations between machines. I think we could add a check of EKEdiss and APEgen back into test_twolayer_qg.

crocha700: Yes. Just updated my comment above. Even in a single machine there is differences between python distributions/ numpy versions, etc.

crocha700: Also, using the ```LayeredModel``` subclass with nz=2 and same config, I get, with anaconda,

```{python}
In [14]: m.get_diagnostic('EKEdiss')*(32**2)
Out[14]: 1.4808828678559061e-07

In [4]: m.get_diagnostic('EKEdiss')*(32**2)
Out[4]: 1.4802470150176962e-07

In [5]: APEgenspec = m.get_diagnostic('APEgenspec')
In [6]: APEgen = 2*APEgenspec
In [7]: APEgen[...,0] = APEgen[...,0]/2.
In [8]: APEgen[...,-1] = APEgen[...,-1]/2.
In [9]: (APEgen.sum()/m.M**2) * (32**2)
Out[9]: 2.5224036997131185e-07
```

crocha700: The differences are actually less than 0.07%, which are indeed very small.

pyqg/tests/test_reference_solns.py
```diff
@@ -0,0 +1,244 @@
...
+        #   appears to be less than 0.1%
+        rtol, atol = 0.001, 0.
+
+        np.testing.assert_allclose(q1norm, 9.561430503712755e-08, rtol=rtol, atol=atol,
+                    err_msg= ' Inconsistent with reference solution')
+
+        diagnostic_results = {
+                'APEgen': 2.5225558013107688e-07  / (m.nx**2),
+                'EKEdiss': 1.4806764171539711e-07 / (m.nx**2),           
+                }
```
rabernat: Wow, lots of stuff happened last night!

I see that ``APEgen`` and ``EKEdiss`` have now made it into the test. This is definitely an improvement.

There are now lots of commented out / skipped diagnostic tests below. Maybe in the comments we should just clarify what they are, where they came from, and whether we plan to fix them or drop them. For our own sanity in the future.

crocha700:
I managed to make all tests pass against Malte's original numbers to within less 1% (most tests actually pass within 0.1%). Towards that end, I implemented a [new diagnostics tool that performs the "spectral sum"](http://git.io/vlNqu) taking into consideration that we're using real FFT. We have to use that as oppose to .sum() when calculating the total sum of a spectral variable in order to satisfy Parseval's theorem (this is similar to `spec_var`, and in fact `spec_var` could leverage on `spec_sum` but that will  wait for the next PR). Other necessary hacks were to compare EKE = EKE1+EKE2 instead of individual components, and some normalization problems.

The normalization issue is not settled, but please let's move that discussion back to #90.

rabernat:
I think this is ready to merge (once landscape does it's thing).

crocha700:
Just to summarize we identified three problems:

1. The ```tavestart``` bug
 That lead to the change to new values (now everything is restored to Malte's original values, a.k.a old values; I restored all old values in f7d8ff3; this PR)
2. `EKEdiss` and `APEgen` were not being correctly calculated (a factor of H2/H) was missing (@MFJansen fixed that in ddd5559)
3. The sum over wavenumbers was not updated when we implemented real FFT (I fixed that in f7d8ff3; this PR).

MFJansen:
Just to clarify:
APEgen was not correctly because of 3 (since it was computed as a spectral sum) - the layer depth ratio was actually accounted for correctly. EKEdiss was wrong for two reasons: one being the spectral sum issue, and the other one being the layer depth issue.

MFJansen:
I'm glad we sorted this out!

I also agree that the normalization for the spectral budget terms still needs to be looked at, so let's keep issue #90 open. I'll try to look into this sometime soon, but definitely won't get to it this week.

crocha700:
Great! I just updated my comment above so that we have a good summary of the changes.

PR #124, opened 2015-11-17
trying to make codecov work

rabernat:


rabernat:
Something is wrong with this. Travis is not building the new commit.

PR #127, opened 2015-11-17
reset travis config

rabernat:
Codecov is still not working. This is lame.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ notifications:
 # Setup anaconda
 before_install:
 # python 3 needs something different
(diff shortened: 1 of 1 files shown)


crocha700:
As I understand, this simply reverts the changes in 4a58e30, right? Should I merge this?

rabernat:
Yes, I'll take care of this. I was just messing around, trying to get test coverage reports going. But it is too much work.

FYI, we need to change the nosetest command to something like the following
```bash
nosetests -v --with-cov --cover-package=pyqg
```

But this doesn't work. For me it just hangs forever.

PR #138, opened 2016-03-24
refactoring travisci build with python 3 tests

rabernat:
A sample build with the new setup:
https://travis-ci.org/rabernat/pyqg/builds/118349708

fixes #130 

updated channel specification
trying with and without pyfftw
hack to skip tests that rely on pyfftw
got rid of unnecessary import test
added missing yaml files
fix setup.py print statements
try to make test skipping work
fixing of setup.py strings for python 3

PR #139, opened 2016-03-25
python 3 compatibility

### LINKED ISSUE #129: Python 3 support
It's long overdue

### LINKED ISSUE #130: Refactor travisci to support matrix builds
This is a necessary step towards #129.

A nice example of how it would work can be found here:
https://github.com/xgcm/xgcm/blob/master/.travis.yml

rabernat:
With this PR, the tests pass in python 2.7 and 3.5. There may still be some residual non-python 3 compliant code floating around, but we won't find it until we improve the tests.

Fixes #129
Fixes #130 

(I think this supercedes #138)

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,12 @@
+[run]
+branch = True
+
+[report]
+exclude_lines =
+    if self.debug:
+    pragma: no cover
+    raise NotImplementedError
+    if __name__ == .__main__.:
+ignore_errors = True
+omit = pyqg/tests/*
+       pyqg/__init__.py
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,43 +1,51 @@
-# based on this
-# https://gist.github.com/dan-blanchard/7045057
+# Based on http://conda.pydata.org/docs/travis.html
 language: python
-sudo: false
-python:
-  - 2.7
-#  - 3.3
+sudo: false # use container based build
 notifications:
   email: false
 
-# Setup anaconda
+matrix:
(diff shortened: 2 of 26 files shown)


rabernat:
@crocha700, as you can see from the [travis-ci build](https://travis-ci.org/pyqg/pyqg/builds/118385837), this works with 2.7 and 3.5, both with and without pyfftw. I found an [anaconda channel](https://anaconda.org/nanshe/pyfftw) that hosts a python3.5-compatible pyfftw.

The only weird thing happening is that py.test doesn't seem to want to exit with python 3.5 and pyfftw, so [the travis-ci build](https://travis-ci.org/pyqg/pyqg/jobs/118385841) is still running, even though the tests complete successfully. I'm not sure how to fix that.

rabernat:
Well I restarted the build and it worked fine. So might have just been a fluke.

We'll have to keep an eye on that issue.

crocha700:
Thanks for making these changes @rabernat !

Could you please update [whatsnew](https://github.com/pyqg/pyqg/blob/master/docs/whats-new.rst) and  mention the anaconda channel you found in [installation](https://github.com/pyqg/pyqg/blob/master/docs/installation.rst)?

rabernat:
Still having the same issue with the python 3.5 build. I wonder if it's a pyfftw issue. Doesn't happen with out pyfftw. 

I can also intermittently reproduce the problem on my laptop.

rabernat:
I'm going to merge this and file a new issue about the hanging problem.

PR #146, opened 2016-04-27
updating to 0.2.0

rabernat:


rabernat:
There is something funky going on with the particle tests. This has got to be a conda issue. Those are the only tests that require scipy (for interpolation).

Issue #144, opened 2016-04-07
Automatic / Adaptive Timestepping


rabernat:
Currently the timestep is set manually. This is both a challenge for novice users (how do you know how to pick) and a major performance bottleneck (if the timestep is unnecessarily small). 

We should make the timestep choice automatic based on a target CFL condition. I found some nice documentation about how to do this with Adams Bashforth.

http://lucan.me.jhu.edu/wiki/index.php/The_flow_solver_implementation#Adaptive_time_stepping
http://lucan.me.jhu.edu/wiki/index.php/Second-order_variable_time_step_Adams-Bashforth_method

rabernat:
@crocha700, @MFJansen -- do you have any thoughts about adaptive timestepping?

@anirban89 and I are doing some simulations where we really don't know what the equilibrated KE will be, and we are having a hard time guessing the right timestep. Too small and it takes forever, too large and CFL blows up.

What have you been doing about this?

crocha700:
Quick-and-dirty solution: start with a fraction (perhaps 1/10?) of the time-scale of the most linearly unstable mode. You can then test the CFL number every once in a while, and reduce the time-step if the CFL number  is too large --- one way to test if this is effective without messing up with the kernel/model class is to use ``run_with_snapshots''.

Long-term-solution: Implement more robust time-steppers and use an automatic approach as you suggest above.

rabernat:
> You can then test the CFL number every once in a while, and reduce the time-step if the CFL number is too large

This is what we had in mind, but the code as is doesn't actually allow you to change the timestep once the kernel is initialized. So we will at least have to implement that.

MFJansen:
I don't have much to add. Yes, an adaptive time step which periodically re-calculates the time step based on a CFL criterion would be a good addition. From an algorithmic point of view this seems pretty straightforward (although as @rabernat points out above you have to be little bit carefull with the A-B time-stepping). I'm obviously a bit behind on the Kernel implementation. Why can't we change the time-step in the Kernel? Isn't it a model attribute?

rabernat:
> Why can't we change the time-step in the Kernel? Isn't it a model attribute?

The kernel uses a hidden attribute called `._dt` to store the timestep, while the python level uses a different variable `.dt`. This is stupid and there is no good reason for it, just me making the code unnecessarily complicated.

PR #156, opened 2016-10-06
Kernel refactor

rabernat:
This is an attempt to roll back some of the unnecessary complexity that got added when I created the cython kernel. There were lots of duplicated attributes between the cython and python levels (e.g. `model.dt` vs. `model._dt`) that were making it hard to understand what is going on (see #144). 

The main thing happening here is the elimination of methods like `_kernel_init()` and `set_qh()` in favor of direct access to the cython-level model attributes. For example, to set the pv, you can now just say

```python
m.q = some_numpy_array
```

Or if you want to change the timestep, you can just do
```python
m.dt = 100
```
(note that this also resets the adams bashforth timestepping to start with a forward-Euler step)

I hope this is the first of several PRs to clean up and streamline the code, making it easier to move forward with the implementation of other new features.

cc: @crocha700

### DIFF
--- a/pyqg/kernel.pyx
+++ b/pyqg/kernel.pyx
@@ -1,7 +1,7 @@
 #cython: profile=True
 #cython: boundscheck=False
 #cython: wraparound=False
-#cython: nonecheck=False 
+#cython: nonecheck=False
 #from __future__ import division
 import numpy as np
 import warnings
@@ -13,10 +13,10 @@ from cython.parallel import prange, threadid
 include '.compile_time_use_pyfftw.pxi'
 IF PYQG_USE_PYFFTW:
     import pyfftw
-    pyfftw.interfaces.cache.enable()  
+    pyfftw.interfaces.cache.enable()
 ELSE:
     import numpy.fft as npfft
-    warnings.warn('No pyfftw detected. Using numpy.fft') 
+    warnings.warn('No pyfftw detected. Using numpy.fft')
 
 # We now need to fix a datatype for our arrays. I've used the variable
 # DTYPE for this, which is assigned to the usual NumPy runtime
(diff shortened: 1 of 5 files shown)


rabernat:
Surprised by a failing test...everything passed on my machine.

rabernat:
There is a single test failure (`test_model.py::PyqgModelTester::test_advection`) related to the presence of a NaN. On my mac, I don't get this error. It doesn't seem important, since all of the other verification tests (including the reference solutions) are fine. Still frustrating to debug, since I can't reproduce it.

rabernat:
I fixed the tests...had to do with how the variables are initialized.

Would love some feedback before merging.

pyqg/tests/test_model.py
```diff
@@ -197,6 +197,8 @@ def test_timestepping(self, rtol=1e-15):
         self.assertEqual(self.m.tc, 0)
         # step forward first time (should use forward Euler)
         self.m._forward_timestep()
+        print('here')
```
crocha700: Is this necessary?

pyqg/kernel.pyx
```diff
@@ -145,14 +156,23 @@ cdef class PseudoSpectralKernel:
         difftout = self._empty_real()
         self._dummy_ifft_out = difftout
 
+        # time stuff
+        self.dt = 0.0
```
rabernat: My approach here was simply to initialize all the variables in the kernel. They can (and should) be overwritten later.

pyqg/tests/test_model.py
rabernat: Oops! Leftover from debugging. Anyway, it was removed in a later commit.

PR #159, opened 2017-01-17
[READY FOR REVIEW] Feature zonal jet

mbueti:
This is an initial attempt to implement a background flow with meridional structure in the QG model

- [x] Changes need to be carried over to surface, barotropic, and layered models
- [x] Tests should be implemented
- [x] Multiple additional examples should be removed and/or consolidated before planning to merge
- [x] Overload the model constructor to accept both constant and structured U, to maintain the current API
- [x] Update linear stability analysis to reflect changes
- [x] Implement nonlinear coupling of zonal background flow with perturbation
- [x] Add viscous coupling to linear stability analysis

Any review, comments, or critiques are greatly appreciated!


MFJansen:
Welcome to pyQG @mbueti ! 
A jet implemented as a background flow will be nice for linearized problems. It can also be useful for fully non-linear simulations, but only if combined with some sort of restoring of the "perturbations". 
Notice that in the full nonlinear model the separation into "background" and "perturbation" is actually sort of artificial. The only way it currently enters is via the periodic boundary conditions -- as the model is periodic in the "perturbations", which enforces a mean PV gradient given by the background shear (and beta). The meridional structure of the background shear however doesn't actually enter here, which you should be able to confirm by comparing two simulations with and without meridional structure in the background flow. The full flow should be (statistically) identical in the two simulations.

If you have a background flow in the lower layer, it will also enter via friction, which only acts on the perturbations -- thus effectively amounting to a restoring to the background flow in the lower layer. To make full use of the variable background flow field, it would be nice to have the option to restore the flow in all layers. It may also be nice to have an option for a restoring (i.e. damping of the perturbations) that only acts on zonal wavenumber zero. That would allow to study the eddy-jet interactions, without damping the eddies.

examples/run-two-layer-jet-only.py
```diff
@@ -0,0 +1,64 @@
...
+    f = plt.contourf(m.q[0,:],levels=clevels, extend='both')
+    plt.contour(m.q[0,:],levels=levels, extend='both', colors='#444444')
+    plt.clim([-0.0005,0.00075])
+    plt.colorbar(f)
+    plt.pause(0.01)
+    plt.draw()
+    plt.savefig('Q_with_jet_' + ('%04d' % i) + '.png', bbox_inches='tight')
+    i += 1
+
+# now the model is done
```
rabernat: Why are there all of these new examples? Are you sure you meant to check them in with your PR?

I am open to adding new examples, but we should discuss which ones and why.

rabernat: Why are there all of these new examples? Are you sure you meant to check them in with your PR?

I am open to adding new examples, but we should discuss which ones and why.

pyqg/kernel.pyx
```diff
@@ -72,9 +72,9 @@ cdef class PseudoSpectralKernel:
     cdef readonly DTYPE_com_t [:] _il
     cdef public DTYPE_real_t [:,:] _k2l2
     # background state constants (functions of z only)
-    cdef DTYPE_real_t [:] Ubg
-    cdef DTYPE_real_t [:] Qy
-    cdef readonly DTYPE_com_t [:, :] _ikQy
+    cdef DTYPE_real_t [:,:] Ubg
+    cdef DTYPE_real_t [:,:] Qy
+    cdef readonly DTYPE_com_t [:, :, :] _ikQy
```
rabernat: Just trying to understand your approach.
It looks like you have added another dimension to the background state, which is exactly what is needed in order to implement this feature correctly.

pyqg/kernel.pyx
```diff
@@ -344,7 +344,7 @@ cdef class PseudoSpectralKernel:
                     # overwrite the tendency, since the forcing gets called after
                     self.dqhdt[k,j,i] = -( self._ik[i] * self.uqh[k,j,i] +
                                     self._il[j] * self.vqh[k,j,i] +
-                                    self._ikQy[k,i] * self.ph[k,j,i] )
+                                    self._ikQy[k,j,i] * self.ph[k,j,i] )
```
rabernat: Looks good...here the advection of the now *3D* background PV gradient gets added to the tendency.

mbueti:
@rabernat sorry, these were included in my examples directory only as a matter of convenience. They should absolutely be pruned out before any consideration for merging is made, since they’re obviously one-off experiments, rather than truly illustrative examples of the model’s functionality!

pyqg/kernel.pyx
```diff
@@ -516,7 +516,7 @@ cdef class PseudoSpectralKernel:
     property ufull:
         def __get__(self):
             return np.asarray(self.u) + \
-                np.asarray(self.Ubg)[:,np.newaxis,np.newaxis]
+                np.expand_dims(np.asarray(self.Ubg),axis=2)
```
rabernat: Can you explain why it is necessary to call `expand_dims`?

pyqg/qg_model.py
```diff
@@ -124,12 +124,13 @@ def _initialize_background(self):
...
-        self.Qy2 = self.beta - self.F2*(self.U1 - self.U2)
+        # need to calculate actual PV gradient
+        self.Qy1 = self.beta + self.F1*(self.U1 - self.U2) + np.gradient(np.gradient(self.U1,self.dy),self.dy)
+        self.Qy2 = self.beta - self.F2*(self.U1 - self.U2) + np.gradient(np.gradient(self.U2,self.dy),self.dy)
         self.Qy = np.array([self.Qy1, self.Qy2])
         # complex versions, multiplied by k, speeds up computations to precompute
-        self.ikQy1 = self.Qy1 * 1j * self.k
-        self.ikQy2 = self.Qy2 * 1j * self.k
+        self.ikQy1 = self.Qy1[:,np.newaxis] * 1j * self.k
+        self.ikQy2 = self.Qy2[:,np.newaxis] * 1j * self.k
```
rabernat: We should sort out this shape stuff. Would be nice not to have to add these new axes. (Not directly related to your PR however.)

mbueti:
Thanks for the notes and suggestions @MFJansen and @rabernat! I'll try to address them as soon as possible. It may take a little time, as Malte knows I'm working this in my spare time while working some other research projects and planning my wedding, so it's going to be a labor of love for the next while!

pyqg/kernel.pyx
mbueti: Absolutely! This comes from needing to essentially needing to tuck a new dimension in the "middle" of the Ubg dimensions, since it can now have structure in both the j and k indexes. 

The bracketed notation using `np.newaxis` does not necessarily play well and give you the shape needed for u and Ubg to be broadcast together. However, `np.expand_dims` lets you arbitrarily choose where you want to pop in your new dimension, and coerce the data into the shape needed.

mbueti:
@MFJansen @rabernat several updates this evening! The ability to add meridional structure to the zonal background current has been implemented in all of the model variants. Along with those changes, the interface on accepting U has been updated. If an array is provided, the model will accept that structure. If a scalar value is, the initialization will coerce it into an array of constant value. This allows the user to specify this structure, but also maintains the existing API, bringing almost all tests back to passing.

The glaring exception here is that the linear stability analysis is blowing up, which is next on my to do list!

rabernat:
Wow, glad to see so much progress on this PR! (And glad I'm not the only one working on a Friday night ;)

mbueti:
Well, when the fiancée's working a 24 hour call...I kinda have no excuse!

Next step's to get some tests in place around the meridional structure itself, and look at implementing some of @MFJansen's suggestions for more complete physics!

pyqg/tests/test_reference_solns.py
```diff
@@ -92,45 +92,45 @@ def test_two_layer(self):
...
-        np.testing.assert_allclose(pnorm, 1493.217664248918, rtol, atol,
-                err_msg= ' Inconsistent with reference solution')
-        np.testing.assert_allclose(ke, 0.9950360837282386, rtol, atol,
-                err_msg= ' Inconsistent with reference solution')
+#         np.testing.assert_allclose(qnorm, 89101.741238768518, rtol, atol,
+#                 err_msg= ' Inconsistent with reference solution')
+#         np.testing.assert_allclose(pnorm, 1493.217664248918, rtol, atol,
+#                 err_msg= ' Inconsistent with reference solution')
+#         np.testing.assert_allclose(ke, 0.9950360837282386, rtol, atol,
+#                 err_msg= ' Inconsistent with reference solution')
```
rabernat: Presumably the commenting out of these tests is just temporary?

You will also eventually want to add a new test with a zonal jet.

mbueti: Exactly right! What I'm digging into right now in fact

rabernat:
Feel free to ping me when you feel this is ready for a full review.

mbueti:
@MFJansen now that the tests are passing and the API is reverse compatible, I'm circling back to coupling the background shear into eddy flow.

Would you envision this as an explicit frictional drag term? As a note, I have updated the background PV shear term so that it does enter the perturbative PV tendency. Thanks!

mbueti:
Also, for anyone using AWS, there's an EC2 instance preconfigured with pyqg and all needed libraries installed with AMI ID ami-15a65703

MFJansen:
@mbueti: just to clarify, when you say you "updated the background PV shear term so that it does enter the perturbative PV tendency" you just mean that advection acts on the full background PV?
That's good and of course how it should be. But as I discussed earlier, the whole separation into "background" and "perturbation" is somewhat artificial in the fully nonlinear model. The only ways it enters is via the meridional (periodic) boundary conditions (although only the meridionally integrated background PV gradient matters here), the initial conditions (although a jet in the initial conditions of course could have been achieved more easily), as well as via the frictional drag. The latter is the most interesting here. The way friction is currently implemented, the drag acts on the "perturbations" in the lower layer, that is we are really restoring the lower layer to the background jet profile. A similar linear restoring could be implemented in all layers to restore to the men jet. Alternatively (or additionally as an extra option) one could introduce a damping of only the zonal-wavenumber 0 perturbations -- that amounts to a restoring of the zonal mean flow to the background profile, without affecting the "eddies".

mbueti:
@MFJansen sorry for the delay getting back on this, I've pushed up another update with a first pass linear viscous restoration in the wave number zero between the eddy and mean flows. Is this the direction you were thinking?

MFJansen:
@mbueti good to hear you are making progress! If I see this correctly the code you implemented only restores wavenumber (0,0), that is zero zonal and zero meridional wavenumber (which actually I believe should have no effect at all). I was thinking about a restoring of the zonal mean flow - that is all components with meridional wavenumber zero. You also seem to be implementing a restoring of the relative vorticity. Instead, I would just have a linear restoring of the PV perturbation, i.e. dqdt=-r q;

mbueti:
@MFJansen that's correct, I had been only having the coupling act against the relative vorticity. I think the PR should now be more in line with what you were suggesting. I had just wanted to play with some test cases to make sure the results looked reasonable. I think the last thing we'd want here, assuming it passes muster, is a configuration flag to turn this coupling on or off

MFJansen:
@mbueti I don't quite understand what you did in your last commit. I think you want "qh" there instead of "Qh". q should by construction be the deviation from your prescribed (zonally symmetric) background flow, so to restore the total flow to that background flow, you simply need to damp q.
It would also be good to define a separate restoring rate for this, rather than using rek (which if I remember correctly is also used for the bottom friction). The restoring can then be turned on simply by setting the restoring rate to a non-zero value.

mbueti:
@MFJansen hm, that's odd, it looks like a pre-commit linter I had hooked in altered the case on that term at some point. I've removed it, and corrected the restoring now. I've also added an independent rbg coefficient as suggested (naming obviously a placeholder for now). The big task left is to integrate the restoring into the linear stability analysis, as with the bottom friction term.

mbueti:
@MFJansen @rabernat okay, I think this is ready for another round of review. The suggestions from Malte should be addressed, and there's some additional testing around the initialization of meridional structure of zonal currents, and that it is accounted for in the mean PV gradient. 

The reference testing needed to have rbg=0 enforced, as having it non-zero alters the perturbation PV so it no longer matches the reference values (as would be expected). Reference testing around these changes is definitely a nice to have!

rabernat:
I just realized that this has been hanging around unmerged for a long time. My apologies!

pyqg/model.py
```diff
@@ -315,7 +319,7 @@ def stability_analysis(self,bottom_friction=False):
 
         Q =  I[:,:,np.newaxis,np.newaxis]*(self.ikQy - self.ilQx).imag
 
-        Uk =(self.Ubg*I)[:,:,np.newaxis,np.newaxis]*self.k
+        Uk =(self.Ubg*I[:,:,np.newaxis])[:,:,:,np.newaxis]*self.k
```
rabernat: I think we need the docstring to be modified somehow to reflect the changes to `Ubg`. Has the API changed? if not, how does one specify the background flow?

mbueti:
@rabernat no worries, and sorry I missed your comments, they actually came in when I had all of my github notices disabled while I was back east getting married! I'll review this, week, thanks!

pyqg/model.py
mbueti: @rabernat I can absolutely update the doctoring and API docs. The API here has been extended. So, we can specify a Ubg via U, or U1/U2 parameters passed into the model that are rank 1 arrays, allowing for meridional structure. I overloaded this parameter in such a way that you can also continue using a scalar/constant background flow, making the updates backwards compatible.

mbueti:
@rabernat just giving this another bump

rabernat:
I'm sorry I lost track of this PR. It is big, and my detailed technical knowledge of how pyqg works has faded a bit.

The tests pass and the code looks clean. At this point. I'm happy to merge. @MFJansen, are you still following this?

MFJansen:
I have not been following this recently. I can try to have a look at it over the weekend. Alternatively, if @mbueti is still in Chicago, maybe we can meet at some point to go over this? 
From just a brief glance over the changes: What are the additions in examples/run_two-layer.py about?

rabernat:
What would be great would be to have a new doc page similar to [this one](http://pyqg.readthedocs.io/en/latest/examples/layered.html) which demonstrates how to use the new features.

mbueti:
I'm happy to put together a demo/doc page for this feature, that's a very reasonable ask @rabernat 

I am still in Chicago through June @MFJansen, and am free to go over the changes made. The two-layer example changes I meant to revert to master, I'd adjusted that during development, before moving my test runs into an .gitignored `experiments` directory. Thanks for catching that!

rabernat:
@mbueti - any update on the possibility of more documentation / example for this new feature.

I would really like to merge this long-standing pull request soon.

mbueti:
@rabernat my apologies! This slipped off of my radar during our relocation back to Rhode Island, and we were away and without internet access last week. The two layer example should now be reverted to master, and I'm going to try setting some time aside to do the documentation over the next week or so. Thanks!

mbueti:
@rabernat a WIP documentation is up on the PR, let me know if this is the direction you were thinking. I'll need to brush up on my jupiter skills for that version, but shouldn't take long.

mbueti:
@rabernat Okay, we should be in business with an example documented!

rabernat:
Great!

Could you just strip the following two files from your PR?
- docs/examples/.ipynb_checkpoints/zonal-jet-checkpoint.ipynb
- tags

mbueti:
@rabernat done and done!

rabernat:
Again, I have forgotten about this PR. LGTM!

PR #161, opened 2017-02-07
replaced some i's with n's for consistency

navidcy:


rabernat:
Ah ok I see there are two PRs...Thanks for this fix!

PR #168, opened 2018-03-09
Update for compatibility with Python 3

pittwolfe:
Dict returns view not list (line 631). Incorrect parens on print statement (line 636).

### DIFF
--- a/pyqg/model.py
+++ b/pyqg/model.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 from __future__ import print_function
 import numpy as np
 from numpy import pi
@@ -626,14 +627,14 @@ class Model(PseudoSpectralKernel):
 
     def describe_diagnostics(self):
         """Print a human-readable summary of the available diagnostics."""
-        diag_names = self.diagnostics.keys()
+        diag_names = list(self.diagnostics.keys())
         diag_names.sort()
         print('NAME               | DESCRIPTION')
         print(80*'-')
         for k in diag_names:
             d = self.diagnostics[k]
-            print('{:<10} | {:<54}').format(
-                 *(k,  d['description']))
+            print('{:<10} | {:<54}'.format(
+                 *(k,  d['description'])))
 
     def _increment_diagnostics(self):
         # compute intermediate quantities needed for some diagnostics
--- a/pyqg/tests/test_diagnostics.py
+++ b/pyqg/tests/test_diagnostics.py
@@ -1,0 +1,17 @@
+from __future__ import print_function
+from future.utils import iteritems
+import unittest
+import numpy as np
+import pyqg
+from pyqg import diagnostic_tools as diag
+
+class DiagnosticsTester(unittest.TestCase):
+
+    def test_describe_diagnostics(self):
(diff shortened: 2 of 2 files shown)


rabernat:
Thanks for this Chris! Unfortunately the `describe_diagnostics` function is not tested by our test suite, so we missed it when adding python 3 support way back in the day. 

Would you feel comfortable adding a basic test of that function? That would help avoid such problems in the future. Doesn't have to be fancy...just call the function to ensure it doesn't error and check that there is some reasonable output.

pittwolfe:
Hi Ryan:

I don’t mind writing a test, but I’m not up on this whole test-based development thing. Can to point me to a basic reference for the proper format of the test and how to get it included in the testing?

Cheers,
Christopher

rabernat:
To run the test suite, you run
```
py.test -v pyqg
```
from the base directory. (You may have to `pip install pytest` first.) This is automatically run by [travis-CI](https://travis-ci.org/pyqg/pyqg) for every pull request.

To add a new test, first look at the existing tests and then add a method to one of the test classes. (The tests are not super well organized or consistent, but at least we have them.) I would recommend adding a method called `test_describe_diagnostics` to `PyqgModelTester` here:
https://github.com/pyqg/pyqg/blob/master/pyqg/tests/test_model.py

rabernat:
Alternatively, you could write a whole new test module called `test_diagnostics.py`. That might be better. We should eventually develop full test coverage for the diagnostics.

pittwolfe:
Added a diagnostics test. Confirms that describe_diagnostics fails with current code but passes with my python3 branch modifications. They’re on two separate branches, so they’ll need to be merged.

Christopher

rabernat:
Hi Christopher! Thanks so much for taking the time to work on this. We sincerely appreciate your contribution.

I would like to clarify how the pull request process typically works for open source projects on github, including pyqg. It is rare for a pull request to be approved straight away in its original form. Most often, the maintainers have some feedback and request some changes (kind of like a paper review). This is what I did in my comment above (https://github.com/pyqg/pyqg/pull/168#issuecomment-373469471). 

The process for addressing these comments is the following:
- Make the requested changes to your local branch (in this case, your local branch is called `python3`)
- Run the test suite again locally
- Push the branch again to origin (`git push python3 origin`)

This will trigger the pull request here on github to be automatically updated and the Travis-CI tests to be run again.

Instead, you opened a new pull request (#169). There is no way to merge two pull requests together from github. And I can't tell a-priori whether they are actually compatible with each other. So it would really be better to have the new test that you wrote in #169 as part is *this* pull request.

To accomplish this, you could either merge the branches locally
- `git checkout python 3; git merge test_diagnostics`
or
- just copy the file `pyqg/tests/test_diagnostics.py` and add it directly to the `python3` branch

Then `git commit` and `git push python3 origin`.

Once you master this process, you will feel confident to contribute to any open source project you wish. pyqg has been a great way for many of us, myself included, to learn about this process, so I hope you see it as an educational experience (rather than a tedious one).

rabernat:
Great! And new we can see that the tests are passing on Travis!

PR #170, opened 2018-03-22
Missing \kappa^2 in the PV equation

biavillas:
Fixes typo in the potential vorticity anomaly equation in Fourier space: missing square on wavenumber \kappa.


rabernat:
Thanks Bia! Good catch!

I'm curious how you are using pyqg for your research.

PR #175, opened 2018-11-15
fixes some typos in layered equations docs

navidcy:
Also, the topographic PV was in both `q_n` and `Q_n`; now it's only part of `Q_n`.

### DIFF
--- a/docs/equations/notation_layered.rst
+++ b/docs/equations/notation_layered.rst
@@ -43,8 +43,8 @@ bottom topography, we obtain the evolution equations
 
    \begin{aligned}
    \label{eq:qg_dynamics}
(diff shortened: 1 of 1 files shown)


rabernat:
We have let this PR lie stale for a long time.

Is anyone else willing to review this? I did not write those docs, so I would prefer to have someone else do this. @cesar-rocha - are you still around?

cesar-rocha:
Yes, I'm still around. 

Good catches, @navidcy. As far as the docs are concerned, the PR seems good to merge, but I'm unsure why all checks have failed. @rabernat, any thoughts?

navidcy:
Could it be related to what @pittwolfe is saying at https://github.com/pyqg/pyqg/pull/182#issuecomment-486790948?

pittwolfe:
No, it's related to what I point out in #189: The advection tests without FFTW always fail. This has apparently been a problem for at least 6 months. @rabernat says the problem isn't with PyQG or with numpy's FFT routines, but with the test itself.

rabernat:
The tests failures were unrelated (fixed in #190).

navidcy:
Just noticed that the changes don't appear in the docs here:
https://pyqg.readthedocs.io/en/latest/equations/notation_layered.html
Are the docs generated by travis?

PR #182, opened 2019-04-25
rbg should default to zero

pittwolfe:
Prior to the updates for the zonal jet example, PyQG had no relaxation to the mean flow. The relaxation is controlled by the parameter `rbg` which defaults to `5.787e-7`. This leads to extra damping relative to the previous version for the same input parameters. In order to recover original behavior, `rbg` should default to zero.

### DIFF
--- a/pyqg/model.py
+++ b/pyqg/model.py
@@ -102,7 +102,7 @@ class Model(PseudoSpectralKernel):
         useAB2=False,               # use second order Adams Bashforth timestepping instead of 3rd
         # friction parameters
         rek=5.787e-7,               # linear drag in lower layer
-        rbg=5.787e-7,               # linear drag in each layer
+        rbg=0,                      # linear drag in each layer
         filterfac=23.6,             # the factor for use in the exponential filter
         # constants
         f = None,                   # coriolis parameter (not necessary for two-layer model


rabernat:
Thanks for reporting this Chris!

It seems like this should have been caught by tests. Is there a test we could add that would cover these sort of parameter defaults?

pittwolfe:
Probably the most straightforward test would be something like MITgcm's testreport system: have a couple of verification cases with known output that can exercise the entire model.

However, it seems like the test system is a little wonky as is. Somehow changing this one parameter has caused PyQG to fail the advection test, which doesn't even involve `rbg`.

rabernat:
I fixed the testing, so we can move forward with this.

My worry here is that we might change a default parameter and break some existing setups. It's hard to know what is the best course of action.

pittwolfe:
In this case, the problem was caused by the **addition** of a new parameter with a non-zero default. That's even harder to test for than changes in pre-existing parameter defaults. The only reason we noticed a problem is because we ran the model to equilibrium and noticed a change in the EKE between model updates. That seems too resource-intensive for routine testing.

pittwolfe:
I still think the background drag should be zero by default since it's not part of the "standard" layered QG setup from textbooks and it's not mentioned in the documentation. It seems extremely likely to trip up any new PyQG user.

rabernat:
@pittwolfe - I hear what you're saying. We made a bad choice early on. But how do we handle the deprecation problem? Imagine you are a pyqg user, you update to a new version, run your simulation, and it blows up because there is no bottom drag. You would be 😡!

The "proper" way to handle such a breaking change is to do a deprecation cycle. Release a version (0.3) with a warning message saying that in version 0.4, the default will change. The problem is that we have not really been doing lots of regular releases of pyqg, so it seems a little silly.

pittwolfe:
This is not about bottom drag, which *is* in the documentation and is carried by the variable `rek`. This is about *background* drag, which relaxes the flow in every layer to the background flow and is carried by the variable `rbg`. This is *not* in the documentation, is not in standard QG, and won't cause the model to blow up if it is zero.

MFJansen:
I'm actually confused why rbg even still exists. This was added as part of PR#159, which I thought we reverted.

rabernat:
Right, thanks for the clarification. Looks like this is another addition from #159 that is still laying around. It's toast.

pittwolfe:
Awesome!

rabernat:
Just for clarification, "reverting" #159 was not trivial, since lots of other (useful) development happened since then. There still seem to be some things around that we want to remove. If you got to a code page (e.g. https://github.com/pyqg/pyqg/blob/master/pyqg/model.py) and click on "blame," you can see them.

If anyone wants to work on doing this cleanup more carefully, it would be most welcome.

rabernat:
Ok, so something is still fishy. I have tried two times to revert the changes from #159, but every time they are merged it doesn't quite work.

rabernat:
I think I finally fixed things by force-pushing the correct branch to master.

Issue #185, opened 2019-05-08
model.py


mnkmishra02:
In model.py using vertical modes the radii[0] is wrong

rabernat:
Hi @mnkmishra02 - thanks for your report, and welcome to the pyqg community! :)

Could you elaborate a bit on this issue? How did you reach this conclusion? Is the theory we are using wrong (https://pyqg.readthedocs.io/en/latest/equations/notation_verticalmodes.html), is this some sort of bug in the code?

rabernat:
Is this the code you are referring to?

https://github.com/pyqg/pyqg/blob/4f41584a12bcbf8657785b8cb310fa5065ecabd1/pyqg/model.py#L247-L248

mnkmishra02:
radii[0] = np.sqrt(self.g*self.H)/np.abs(self.f)

rabernat:
Hi @mnkmishra02 - thanks for your reply to my question! :)

I'm not sure I understand your response. What does the expression
```
radii[0] = np.sqrt(self.g*self.H)/np.abs(self.f)
```
mean in this context? Is this what you think should be the correct value of radii for the model with vertical modes?

In general, if you could be a bit more verbose it would help our communication to converge more quickly.

MFJansen:
Why was this closed? 

I'm not really familiar with this part of the code, but it looks to me like @mnkmishra02 is correct: radii[0] should simply be the barotropic deformation radius, i.e. radii[0] = np.sqrt(self.g*self.H)/np.abs(self.f).

The same issue has also been pointed out by @benmayersohn in issue#178. If whoever wrote this piece of the code can confirm that we are not missing anything and make the fix, that would be great!

pittwolfe:
I have created #186 to address this.

PR #186, opened 2019-05-17
Fixed definition of barotropic deformation radius.

pittwolfe:
As mentioned in #185, the definition of the barotropic deformation radius is run. Should be sqrt(g H)/|f|, was g H/|f|, which doesn't even have the right units.

### DIFF
--- a/pyqg/model.py
+++ b/pyqg/model.py
@@ -244,7 +244,7 @@ class Model(PseudoSpectralKernel):
         # deformation wavenumbers and radii
         self.kdi2 = evals[asort]
         self.radii = np.zeros_like(self.kdi2)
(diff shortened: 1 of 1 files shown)


cesar-rocha:
Thanks for addressing this.

 The test test_vertical_modes.py. Could you also fix that?

pittwolfe:
Note that this failed the `radii` check in Travis because the original value is wrong. The barotropic deformation radius has been incorrect since it was first introduced in #108 four years ago. (Hope nobody's results depended on that!)

pittwolfe:
I’m not sure I understand. Doesn’t letting me change the test my pull request fails basically negate the point of having tests? 

Christopher

MFJansen:
I accepted this and also fixed the test_vertical_modes.py test, which had the same error (which is why fixing the code caused a failed test)

PR #187, opened 2019-05-17
Super of layered model overwrites g

pittwolfe:
The superclass of LayeredModel has `g` has a keyword argument, so LayeredModel either shouldn't or should pass `g` to `super.__init__` directly. Otherwise the value of `g` passed while instantiating LayeredModel will get overwritten by `Model`.


pittwolfe:
Oops. Pushed more changes than I was intending.

PR #193, opened 2019-06-11
Ubg Updates Fixes #192

### LINKED ISSUE #192: Error running LayeredModel
I've run into two errors when running the LayeredModel, and both appear to be linked to array sizes with zonal jet feature #159.   @mbueti, can you look into this?

The first error is related to the size of Vbg in the APEgenspec diagnostic in layered_model.  Looks like Vbg has the wrong size.  This is triggered when a run passes tavestart.

The second error is layered_model _calc_cfl.  It looks like Ubg has an extra np.newaxis.  This is triggered when a run passes twrite.

These errors can be generated by the 3-layer baroclinic instability code in the documentation, but I'll copy it here anyway.  Setting twrite=1 should show the second error.

```python
import numpy as np
import pyqg
from matplotlib import pyplot as plt

L =  1000.e3     # length scale of box    [m]
Ld = 15.e3       # deformation scale      [m]
kd = 1./Ld       # deformation wavenumber [m^-1]
Nx = 64          # number of grid points

H1 = 500.        # layer 1 thickness  [m]
H2 = 1750.       # layer 2
H3 = 1750.       # layer 3

U1 = 0.05          # layer 1 zonal velocity [m/s]
U2 = 0.025         # layer 2
U3 = 0.00          # layer 3

rho1 = 1025.
rho2 = 1025.275
rho3 = 1025.640

rek = 1.e-7       # linear bottom drag coeff.  [s^-1]
f0  = 0.0001236812857687059 # coriolis param [s^-1]
beta = 1.2130692965249345e-11 # planetary vorticity gradient [m^-1 s^-1]

Ti = Ld/(abs(U1))  # estimate of most unstable e-folding time scale [s]
dt = Ti/200.   # time-step [s]
tmax = 20*Ti      # simulation time [s]

m = pyqg.LayeredModel(nx=Nx, nz=3, U = [U1,U2,U3],V = [0.,0.,0.],L=L,f=f0,beta=beta,
                         H = [H1,H2,H3], rho=[rho1,rho2,rho3],rek=rek,
                        dt=dt,tmax=tmax, twrite=5000, tavestart=Ti*10, ntd=1)

sig = 1.e-7
qi = sig*np.vstack([np.random.randn(m.nx,m.ny)[np.newaxis,],
                    np.random.randn(m.nx,m.ny)[np.newaxis,],
                    np.random.randn(m.nx,m.ny)[np.newaxis,]])
m.set_q(qi)

m.run()
```

mbueti:


MFJansen:
I'm not sure if this is actually right physically if Ubg and Vbg are not constant (looking at the APEgenspec calculation here, although there may be issues elsewhere as well). The equations in the documentation assume Ubg and Vbg are constants.

mbueti:
@MFJansen mmmm, that's a great point. In this case, Ubg and Vbg are constant in each layer, but we encountered an issue of dimensionality, as Ubg is now defined as uniform over a 1 dimensional array if no structure is provided.

Given that situation, we have two ways forward, let's call them the "right way" and the "fast way".

The right way is obviously to reformulate all of the diagnostics to account for non-uniform Ubg in any given layer. Reasonable, but will just take some dedicated time by somebody.

In the meantime, I'd advocate for the fast way as a stopgap, that being to throw a friendly exception if a diagnostic that is incompatible with horizontally nonuniform flow is requested in conjunction with structured Ubg. This keeps people from deriving incorrect answers, and leaves the option open for us to reformulate at will.

mbueti:
Whether we reformulate now or later, I would definitely advocate for fixing this issue more immediately, and spinning those off as their own issues to be addressed :-)

MFJansen:
Sounds good. If you can you implement that stopgap, and put an issue in that this should be dealt with at some point, I'm happy to merge this.

mbueti:
@MFJansen updated, and added a new issue at https://github.com/pyqg/pyqg/issues/196. The only outstanding question is whether an exception should be thrown when attempting to access other APE diagnostics? It's very easy to do so, if that's desirable!

worthamc:
@mbueti thanks for fixing these.  Your APEgenspec fix does correct the dimensions, but it looks like it introduces two singleton dimensions.  I think the code below gives the same result, but without the new dimensions.

                    function =(lambda self: (self.Hi[:,np.newaxis,np.newaxis]*
                                (self.Ubg[:,:,np.newaxis]*self.k[np.newaxis,:,:] +
                                 self.Vbg[:,np.newaxis,np.newaxis]*self.l)*
                                (1j*self.ph.conj()*self.Sph).real).sum(axis=0)/self.H))

mbueti:
Nice catch @worthamc, this should do us

MFJansen:
I'm closing this, since development of variable Ubg feature should for now continue on a separate branch.

Issue #181, opened 2019-02-12
Operator error for time-mean diagnostics


tlhsieh:
In _increment_diagnostics() in model.py, the += operator doesn't seem to be doing the right thing. Replacing += by explicit assignment solves the problem, e.g., [https://github.com/tlhsieh/pyqg/commit/d0fd5669d30ad9870b38fab9e589df63c1015c33](url)

Here's an example that illustrates the problem. Using the same setup as [https://github.com/pyqg/pyqg/blob/master/docs/examples/two-layer.ipynb](url), m.get_diagnostic('q')[0] would look like 
![unknown](https://user-images.githubusercontent.com/35002088/52664513-372c3080-2ed7-11e9-9475-50debb0035e0.png)

But it is in fact the last snapshot divided by the count. The last snapshot m.q[0] looks like 
![unknown1](https://user-images.githubusercontent.com/35002088/52664530-3eebd500-2ed7-11e9-88e7-24c552afab04.png)

After the change referenced above, m.get_diagnostic('q')[0] would look like 
![unknown2](https://user-images.githubusercontent.com/35002088/52664585-575bef80-2ed7-11e9-93e9-01120a9fc8b3.png)

The example is generated using this notebook [https://github.com/tlhsieh/pyqg/blob/diagnostic_tool/docs/examples/two-layer_diagnostics.ipynb](url)

My python environment follows [https://github.com/pyqg/pyqg/blob/master/ci/environment-py35-pyfftw.yml](url)

rabernat:
Hi @tlhsieh, thanks for reporting this error and sorry no one ever replied to you.

I see you opened a PR in #183 to fix this. Why did you close that?

rabernat:
FYI, this would have been easier to catch if we had proper tests for diagnostics (issue #42).

tlhsieh:
A new PR #201 created, which excludes the unnecessary notebook file.

PR #201, opened 2019-06-27
fixes += error in _increment_diag

tlhsieh:
Fixes an operator error in the diagnostic tool described in #181


pyqg/model.py
```diff
@@ -651,7 +651,7 @@ def _increment_diagnostics(self):
                 if self.diagnostics[dname]['count']==0:
                     self.diagnostics[dname]['value'] = res
                 else:
-                    self.diagnostics[dname]['value'] += res
+                    self.diagnostics[dname]['value'] = self.diagnostics[dname]['value'] + res
```
rabernat: Isn't this the exact same thing?

tlhsieh: I was surprised that they turned out differently as well. As an example, I created two Binder environments: [the first one](https://mybinder.org/v2/gh/tlhsieh/pyqg/87152809b6ce7c90ade48c1cf29f75fa8ad3d95a) built from the master pyqg source ([yml here](https://github.com/tlhsieh/pyqg/blob/binder_github/environment.yml)), and [the other one](https://mybinder.org/v2/gh/tlhsieh/pyqg/aa00057cfb98c8ba662bcf58c499c2e9d671903c) built with this current modification ([yml here](https://github.com/tlhsieh/pyqg/blob/binder_tlh/environment.yml)). The diagnostic tool no longer returns the last snapshot.

asross:
@tlhsieh I haven't noticed this issue locally -- diagnostics seem to be time-averaging correctly, at least on the installs I've tested. I'll plan to close this, but if you can add a failing test case, we can reopen and debug!

PR #205, opened 2019-11-18
Revert zonal jet

### LINKED ISSUE #202: error in spectral divergence term of do_advection
There is an error with ikQy in the spectral divergence part of _do_advection.  The spectral divergence is computed in spectral space, but the ikQy term is in physical space.  This doesn't matter for spatially uniform Ubg, but with a zonal jet this will be wrong.  

I think @navidcy said he has some notes on the proper term here.

rabernat:
This PR reverts the changes in #159, which have resulted in a few problems.

Fixes #202.

I have cherry picked the subsequent commits. The old code is archived in the `zonal-jet` branch.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -5,7 +5,7 @@
 pyqg: Python Quasigeostrophic Model
 ===================================
 
-|DOI| |pypi| |Build Status| |codecov| |docs| |landscape| |binder|
+|DOI| |pypi| |Build Status| |codecov| |docs| |binder|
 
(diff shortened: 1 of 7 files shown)


rabernat:
Don't know why travis is not building this (or any other recent PRs).

rabernat:
I think the problem here is that we merged this with master, but what we really wanted was just to overwrite master.

I'm going to try a different fix.

PR #206, opened 2019-11-20
Shape bug fix

houssamyassin:
In its current state, many of the model configurations cannot be run at all. If you try to run the examples found in the documentation, you will persistently get shape errors. I believe these errors originate  from an attempt to allow the prescribed zonal velocity to depend on the meridional direction (i.e. U=U(y)).  What I did here to remove this feature from the code 'manually'. With these changes, I was able to run all pyqg in all of its configurations (with the exception of the 'zonal jet feature') and reproduce the plots in the examples provided in the documentation.

An alternative to accepting this pull request would be to roll the model back to an earlier commit before the U=U(y) feature was added.


MFJansen:
Thanks @houssamyassin!
We rolled back the merger of the non-constant background flow, which needs further development.  Please confirm that the problems that you encountered are now fixed and that this makes this PR unnecessary.

houssamyassin:
I just tried the layered model example (the one in the documentation) with the roll back and I am still getting the shape errors. 

If you look at (say) lines 168-171 or line 188 in layered_model.py then you'll see that the dimensionality of U is expanded to account for meridional variability. Or if you check lines 77-78 in kernel.py then U and Qy remain two dimensional.

Were these lines present before the the non-constant background flow were added? Those are some of the lines I had to modify to get the variable shapes right. The roll back needs to far enough back so that, in lines 77-78 in kernel.py, U and Qy are one dimensional.

rabernat:
Hi @houssamyassin - thanks for looking into this!

FYI, you can directly quote code right here by doing "copy permalink" from any code page. For example
https://github.com/pyqg/pyqg/blob/f018791d8c18d4228ec38f1cccc55b84063aabd4/pyqg/layered_model.py#L168-L171

You are right, it looks like we haven't quite gone far enough. I thought I had in #205 but apparently not.

rabernat:
@houssamyassin - thanks to you we realized that #205 had was not done properly. I believe #207 does what we want.

I would really appreciate if you could give that a quick review! 😄

houssamyassin:
@rabernat : I must have done something wrong last night when cloning the model. The model is building fine now.

But has the rollback taken effect? If I you look at the current version of the code in the master branch, you'll see

https://github.com/pyqg/pyqg/blob/b2fc0a58544e914bffd139369cc5f3e2121dfe7a/pyqg/layered_model.py#L168-L171
so that Ubg and Qy remain two-dimensional despite the fact that #207 was supposed to remove these lines. 

When I clone the current version of the model, I am still getting the shape errors when I run the examples in the documentation. I think #207 is making the required code changes to get the shapes right but I am not seeing these changes in the current version of the code. Am I missing something? Why doesn't the master branch show the changes in merge #207?

MFJansen:
@rabernat yeah, I just noticed the same thing. Somehow we still haven't successfully reverted the zonal jet feature (which also explains why "rbg" was still around)

rabernat:
Ok I have now force-pushed the correct branch from my laptop to master. Everything should _finally_ be resolved.

rabernat:
Relevant code in question:

https://github.com/pyqg/pyqg/blob/9b8cd508c6cf5b441e2730dae0991e55200900fe/pyqg/layered_model.py#L117-L120

houssamyassin:
Great! All of the examples in the documentation can now be reproduced again. Thanks for doing this!

PR #207, opened 2019-11-20
Revert again with proper merging

rabernat:
To create this branch, I followed these instructions.
https://stackoverflow.com/questions/2862590/how-to-replace-master-branch-in-git-entirely-from-another-branch

I think it does what we wanted in #205.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -5,7 +5,7 @@
 pyqg: Python Quasigeostrophic Model
 ===================================
 
-|DOI| |pypi| |Build Status| |codecov| |docs| |landscape| |binder|
+|DOI| |pypi| |Build Status| |codecov| |docs| |binder|
 
 pyqg is a python solver for quasigeostrophic systems. Quasigeostophic
 equations are an approximation to the full fluid equations of motion in
--- a/pyqg/tests/test_vertical_modes.py
+++ b/pyqg/tests/test_vertical_modes.py
@@ -27,7 +27,7 @@ class VerticalModesTester(unittest.TestCase):
         """ Check deformation radii are computed correctly """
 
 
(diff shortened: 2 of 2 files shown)


pittwolfe:
I am highly supportive of this reversion.

houssamyassin:
Just tried out this branch. I am having trouble compiling the model. When I run `python setup.py install` I get back 
>  'pyqg/kernel' is not a valid module name

and the error message points towards the `setup.py` file. In the newer versions of the code, the module name is modified to `pyqg.kernal` in the `setup.py` file. If I make this change and try to compile again, I obtain error messages

> clang: error: unsupported option '-fopenmp'
> error: command 'gcc' failed with exit status 1

which occur during the `gcc` step in the `setup.py` file. 

The second issue could possibly be machine specific (maybe I have the wrong version of C), but I think the first issue will prevent compilation.

rabernat:
> Just tried out this branch. I am having trouble compiling the model.

@houssamyassin - it seems unlikely that you build problems are related to these specific changes. This PR doesn't touch setup.py, and none of the module names are changed here.

https://github.com/pyqg/pyqg/blob/f018791d8c18d4228ec38f1cccc55b84063aabd4/setup.py#L133-L138

As you can see from the [tests](https://travis-ci.org/pyqg/pyqg/jobs/614716667#L432), this PR builds fine on travis. 

> The second issue could possibly be machine specific (maybe I have the wrong version of C), but I think the first issue will prevent compilation.

We have a check in `setup.py` to see whether openmp is installed:

https://github.com/pyqg/pyqg/blob/f018791d8c18d4228ec38f1cccc55b84063aabd4/setup.py#L115-L119

I'm not sure why this isn't working for you, but again, this PR doesn't change any of that.

I think I'm going to go ahead with merging this. We can try to sort out your issues in a separate issue.

PR #209, opened 2019-11-21
Revert #159

MFJansen:
Supposedly this should revert #159 (although apparently that's not as easy as it looks....)


rabernat:
Malte I think I fixed things through a different method.

MFJansen:
Ok, looks like you've been successful, so I'l close this

PR #214, opened 2019-11-23
[WIP] require pyfftw

rabernat:
Now that pyfftw is easily installed from conda, I think we can just require it as a dependency. This allows us to simplify the kernel.

### DIFF
--- a/pyqg/kernel.pyx
+++ b/pyqg/kernel.pyx
@@ -9,14 +9,8 @@ import cython
 cimport numpy as np
 from cython.parallel import prange, threadid
 
-# see if we got a compile time flag
-include '.compile_time_use_pyfftw.pxi'
-IF PYQG_USE_PYFFTW:
-    import pyfftw
-    pyfftw.interfaces.cache.enable()
-ELSE:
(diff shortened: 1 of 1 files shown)


MFJansen:
I'm personally not a big fan of this. Not everyone uses conda, and it would be nice to make sure that students can easily download and use pyqg no matter what python installation they are using. (Just my two cents. If others agree that this is the way we want to go, I won't make a scene...)

rabernat:
I don't propose merging this right away. It's part of a larger discussion #215, which would involve writing other kernels, including a pure numpy one.

I released pyqg v0.3.0 over the weekend, which should have all of our latest fixes / rollbacks. That  is a stable version which people can continue to use as you describe.

MFJansen:
Sounds good! Thanks!

rabernat:
Note that we are now requiring pyfftw as part of the packaging update in #278.

asross:
It should still be possible to run pyqg without it if you install manually, but pyfftw is now the default for a pip install, which seems to work pretty smoothly. Probably makes sense to close this PR for now :)

PR #221, opened 2020-05-12
Update setup.cfg

mengcz13:
Added "inplace" flag for build_ext when run `python setup.py install`. Solved the issue "ImportError: No module named kernel".


rabernat:
Thanks for this suggestion!

PR #227, opened 2021-03-09
Docs notebook

cspencerjones:
I would like to compile the docs directly from notebooks, rather than via rst files.


rabernat:
This is a great Spencer!

However, we also need to confirm that this configuration can build on read-the-docs. To do that, it would be great to have the [readthedocs GitHub](https://docs.readthedocs.io/en/stable/webhooks.html) integration set up so that we could get a test build from RTD directly on the PR. @jbusecke has done that before and perhaps can advise.

rabernat:
Ok just tried it. @cspencerjones can you push another commit to trigger the build.

You could add nbsphinx to the docs/environment.yml, as I'm pretty sure this will be needed for it to work

https://github.com/pyqg/pyqg/blob/master/docs/environment.yml

cspencerjones:
nbsphinx was already in docs/environment.yml! So I added some spacing instead.

jbusecke:
The [RTD preview](https://pyqg--227.org.readthedocs.build/en/227/) looks good to me. Is this ready to merge?

PR #225, opened 2021-03-09
Start GHA

jbusecke:
This PR is aiming to transition the CI to github actions away from travis.

PR #228, opened 2021-03-10
A simpler GHA test

jbusecke:
Due to some problems I encountered I wanted to test a much simpler workflow (template provided by GH) here.


jbusecke:
OK this works, meaning there is an issue with #225, Closing.

Issue #145, opened 2016-04-07
Wrap model variables with xarray


rabernat:
It would be awesome to wrap the model variables with [xarray](http://xarray.pydata.org/en/stable/) data structures. This would make so many things easier for the user, including plotting and serialization to netCDF.

This would probably be a pretty major refactor.

rabernat:
This would enable #9.

rabernat:
Some more guidance about how this might work...

We should perhaps attach a method called `to_dataset()` to the pyqg `Model` object. The method will return an xarray dataset with all of the model prognostic variables and diagnostic variables in it.

### Dimensions & Coordinates

Every Xarray dataset needs to define its dimensions and coordinates. Pyqg has both spatial coordinates (`x`, `y`) and spectral ones (`k` and `l`), plus `level`. Variables are either spectral or spatial.

### Prognostic variables

The prognostic variables are those that are part of the kernel. They are the non-private variables enumerated here:

https://github.com/pyqg/pyqg/blob/master/pyqg/kernel.pyx#L32-L77

For each variable, we need to identify 
- its dims
- the data itself (easy)
- the attrs. We should add attributes like `units` and `standard_name` and / or `long_name` to the attrs, following [CF conventions](https://cfconventions.org/).

### Diagnostic variables

The diagnostic variables are messier.

https://github.com/pyqg/pyqg/blob/master/pyqg/model.py#L571-L597

We will need to figure out how to encode their dims and attrs. That could be done as a second pass, after first doing the prognostic vars.

rabernat:
Creating a dataarray:

http://xarray.pydata.org/en/stable/data-structures.html#creating-a-dataarray

rabernat:
Some pseudo code I wrote today

```python
spatial_dims = ('lev', 'y', 'x')
spectral_dims = ('lev', 'l', 'k')
dim_database = {
    'q': spatial_dims,
    'u': spatial_dims,
    'qh': spectral_dims,
    'ph': spectral_dims
    
}

variables = {}
for vname in dim_database:
    data = getattr(model, vname)
    variables[vname] = (dims, data)
    
ds = xr.Dataset(variables)
```

hscannell:
This is awesome @paigem! Since we're working asynchronously, I can bite off a bit more work on this today.

hscannell:
[...]

PR #229, opened 2021-03-12
Wrap xarray

hscannell:
This PR address the [issue](https://github.com/pyqg/pyqg/issues/145) of wrapping model variables with xarray. A new function `model_to_dataset` converts the pyqg model output into an xarray dataset. The method `to_dataset` in `model.py` can be called on the model itself to produce the dataset.


hscannell:
We should call this method in the example notebooks.

rabernat:
Thanks for getting this started Hillary! Great work!

It looks like you have added files your branch very liberally. As you can see in the [files changed tab](https://github.com/pyqg/pyqg/pull/229/files), this PR touches 18 files. Many of them are from `.ipynb_checkpoints` folders, which are hidden directories created by jupyter. These don't belong in the pygq repo. In generally, you should not run commands like `git add *`. Only add the specific files you want to incldue in the PR.

The easiest thing is probably to close this and start over with a fresh branch.

PR #230, opened 2021-03-12
wrap model output with xarray

hscannell:
This PR address the [issue](https://github.com/pyqg/pyqg/issues/145) of wrapping model variables with xarray. A new function `model_to_dataset` converts the pyqg model output into an xarray dataset. The method `to_dataset` in `model.py` can be called on the model itself to produce the dataset.

**This is a clean branch with only the two files updated added via commits. Thanks @rabernat for pointing this out ;)**

### DIFF
--- a/pyqg/model.py
+++ b/pyqg/model.py
@@ -673,3 +673,8 @@ class Model(PseudoSpectralKernel):
         warnings.warn("Method deprecated. Set model.q directly instead. ",
             DeprecationWarning)
         self.q = q
+
+    def to_dataset(self):
+        '''Convert outputs from model to an xarray dataset'''
+        from .xarray_output import model_to_dataset
+        return model_to_dataset(self)
--- a/pyqg/model.py
+++ b/pyqg/model.py
@@ -675,6 +675,11 @@ class Model(PseudoSpectralKernel):
         self.q = q
 
     def to_dataset(self):
-        '''Convert outputs from model to an xarray dataset'''
+        """Convert outputs from model to an xarray dataset
+        
+        Returns
+        -------
+        ds : xarray.Dataset
+        """
         from .xarray_output import model_to_dataset
         return model_to_dataset(self)
--- a/pyqg/tests/test_xarray_output.py
+++ b/pyqg/tests/test_xarray_output.py
@@ -1,0 +1,13 @@
+from __future__ import print_function
+import numpy as np
+import pyqg
+import xarray as xr
+
+def test_xarray():
+    m = pyqg.QGModel(1)
+    ds = m.to_dataset()
+    
+    assert type(ds) == xr.Dataset
+
+if __name__ == "__main__":
+    test_xarray()
--- a/pyqg/tests/test_xarray_output.py
+++ b/pyqg/tests/test_xarray_output.py
@@ -8,6 +8,9 @@ def test_xarray():
     ds = m.to_dataset()
     
     assert type(ds) == xr.Dataset
-
-if __name__ == "__main__":
-    test_xarray()
+    
+    expected_vars = ['q', 'u', 'v', 'ph', 'Qy']  
+    for v in expected_vars:
+        assert v in ds
+    
+    
--- a/pyqg/xarray_output.py
+++ b/pyqg/xarray_output.py
@@ -1,154 +1,149 @@
 import numpy as np
 import xarray as xr
 
+# Define dict for variable dimensions
+spatial_dims = ('time','lev','y','x')
+spectral_dims = ('time','lev','l','k')
+dim_database = {
+    'q': spatial_dims,
+    'u': spatial_dims,
+    'v': spatial_dims,
+    'ufull': spatial_dims,
+    'vfull': spatial_dims, 
+    'qh': spectral_dims,
+    'uh': spectral_dims,
+    'vh': spectral_dims,
+    'ph': spectral_dims, 
+    'Ubg': ('lev'),
+    'Qy': ('lev'),
+}
+
+# dict for variable dimensions
+var_attr_database = {
+    'q': {'long_name': 'potential vorticity in real space', 'units': 'meters squared Kelvin per second per kilogram',},
+    'u': {'long_name': 'zonal velocity anomaly', 'units': 'meters per second',},
+    'v': {'long_name': 'meridional velocity anomaly', 'units': 'meters per second',},
+    'ufull': {'long_name': 'zonal full velocities in real space', 'units': 'meters per second',},
+    'vfull': {'long_name': 'meridional full velocities in real space', 'units': 'meters per second',},
+    'qh': {'long_name': 'potential vorticity in spectral space', 'units': 'meters squared Kelvin per second per kilogram',},
+    'uh': {'long_name': 'zonal velocity anomaly in spectral space', 'units': 'meters per second',},
+    'vh': {'long_name': 'meridional velocity anomaly in spectral space', 'units': 'meters per second',},
+    'ph': {'long_name': 'streamfunction in spectral space', 'units': 'meters squared per second',},
+    'Ubg': {'long_name': 'background zonal velocity', 'units': 'meters per second',},
+    'Qy': {'long_name': 'background potential vorticity gradient', 'units': 'meters squared Kelvin per second per kilogram',} ,
+ 
+}
+
+# dict for coordinate dimensions
+coord_database = {
+    'time': ('time'),
+    'lev': ('lev'),
+    'x': ('x'),
+    'y': ('y'),
+    'l': ('l'),
+    'k': ('k'),
+    'nx': (),
+    'ny': (),
+    'nz': (),
+    'nl': (),
+    'nk': (),
+    'rek': (),
+    'tc': (),
+    'dt': (),
+    'L': (),
+    'W': (),
+    'filterfac': (),
+    'twrite': (),
+    'tmax': (),
+    'tavestart': (),
+    'tsnapstart': (),
+    'taveint': (),
+    'tsnapint': (),
+    'ntd': (),
+    'pmodes': (),
+    'radii': (),
+}
+
+# dict for coordinate attributes 
+coord_attr_database = {
+    'time': {'long_name': 'model time', 'units': 'seconds',},
+    'lev': {'long_name': 'vertical levels',},
+    'x': {'long_name': 'real space grid points in the x direction', 'units': 'grid point',},
+    'y': {'long_name': 'real space grid points in the y direction', 'units': 'grid point',},
+    'l': {'long_name': 'spectal space grid points in the l direction', 'units': 'meridional wavenumber',},
+    'k': {'long_name': 'spectal space grid points in the k direction', 'units': 'zonal wavenumber',},
+    'nx': {'long_name': 'number of real space grid points in x direction',},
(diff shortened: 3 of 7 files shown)


pyqg/model.py
```diff
@@ -673,3 +673,8 @@ def set_q(self, q):
         warnings.warn("Method deprecated. Set model.q directly instead. ",
             DeprecationWarning)
         self.q = q
+
+    def to_dataset(self):
+        '''Convert outputs from model to an xarray dataset'''
```
rabernat, line 678:
```suggestion
        """Convert outputs from model to an xarray dataset
        
        Returns
        -------
        ds : xarray.Dataset
        """
```

pyqg/xarray_output.py
```diff
@@ -0,0 +1,152 @@
...
+
+    # dict for dataset attributes
+    ds_attr_database = {
+        'title': 'pyqg: Python Quasigeostrophic Model',
+        'institution': '',
+        'source': ('version: {}'.format(pyqg.__version__)),
+        'history': '',
+        'references': 'https://pyqg.readthedocs.io/en/latest/index.html',
+        'comment': '', 
+    }
```
rabernat, line 109: All of these databases can me moved _outside_ the function definition. (They are essentially "constants" which don't depend on the input.) This will help keep the function itself more concise.

pyqg/xarray_output.py
```diff
@@ -0,0 +1,152 @@
...
+
+
+    # Create list of variables
+    variables = {}
+    for vname in dim_database:
+        data = getattr(m,vname)
+        if 'time' in dim_database[vname]:
+            variables[vname] = (dim_database[vname],data[np.newaxis,...])
```
rabernat, line 117: 👍  perfect solution to this

pyqg/xarray_output.py
```diff
@@ -0,0 +1,152 @@
...
+        'uh': spectral_dims,
+        'vh': spectral_dims,
+        'ph': spectral_dims, 
+        'Ubg': ('z'),
+        'Qy': ('z'),
+    }
+
+    # dict for variable dimensions
+    var_attr_database = {
+        'q': {'long_name': 'potential vorticity in real space', 'units': 'm\u00B2 s\u207B\u00B9 K kg\u207B\u00B9',},
```
rabernat, line 26: What are these symbols? Is this some kind of encoding?

Let's use [CF UDUNITS conventions)[https://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/build/ch03.html} for the units.

pyqg/xarray_output.py
```diff
@@ -0,0 +1,152 @@
...
+        'tsnapint': (),
+        'ntd': (),
+        'pmodes': (),
+        'radii': (),
+    }
+
+    # dict for coordinate attributes 
+    coord_attr_database = {
+        'time': {'long_name': 'model time', 'units': 'seconds',},
+        'z': {'long_name': 'vertical levels', 'units': 'none',},
```
rabernat, line 74: Let's call this `lev` instead of z.

pyqg/xarray_output.py
```diff
@@ -0,0 +1,152 @@
...
+        data = getattr(m,vname)
+        if 'time' in dim_database[vname]:
+            variables[vname] = (dim_database[vname],data[np.newaxis,...])
+        else:
+            variables[vname] = (dim_database[vname],data)
+
+    # Create list of coordinates
+    coordinates = {}
+    cname_to_1D = ['time','z','x','y','l','k']
+    coords_1D = [(np.array([m.t])), (np.arange(1,m.nz+1)), (m.x[0,:]), (m.y[:,0]), (m.l[:,0]), (m.k[0,:])]
```
rabernat, line 124: Let's see if we can avoid hard coding these transformations. Aren't the 1D versions of the arrays available somewhere as part of `m`?

pyqg/xarray_output.py
```diff
@@ -0,0 +1,152 @@
...
+    for cname in coord_database:
+        if cname in cname_to_1D:
+            index = cname_to_1D.index(cname)
+            coordinates[cname] = coords_1D[index]
+        else:
+            try:
+                data = getattr(m,cname)
+                coordinates[cname] = (coord_database[cname],data )
+            except:
+                pass
```
rabernat, line 134: I'm not a fan of this try / except block. Can we refactor a code to avoid needing to do it this way? Why would we ever raise an exception?

pyqg/xarray_output.py
```diff
@@ -0,0 +1,152 @@
...
+
+    # Assign attributes to coordinates
+    for caname in coord_attr_database:
+        if caname in ds.coords:
+            ds.coords[caname].attrs = coord_attr_database[caname]
+
+    # Assign attributes to variables
+    for vaname in var_attr_database:
+        if vaname in ds.data_vars:
+            ds.data_vars[vaname].attrs = var_attr_database[vaname]
```
rabernat, line 149: Let's move these loops into the loops above so we only have to do each loop (vars and coords) once.

rabernat (changes requested):
Great work Hillary! 💪 

I have a few suggestions on your code.

Also, we need TESTS! 🤓  We can talk about that our meeting today.

pyqg/xarray_output.py
```diff
@@ -0,0 +1,152 @@
...
+    coordinates = {}
+    cname_to_1D = ['time','z','x','y','l','k']
+    coords_1D = [(np.array([m.t])), (np.arange(1,m.nz+1)), (m.x[0,:]), (m.y[:,0]), (m.l[:,0]), (m.k[0,:])] 
+    for cname in coord_database:
+        if cname in cname_to_1D:
+            index = cname_to_1D.index(cname)
+            coordinates[cname] = coords_1D[index]
+        else:
+            try:
+                data = getattr(m,cname)
```
rabernat, line 131:
```suggestion
                data = getattr(m, cname, None)
                if data:
                    # do something
```

pyqg/xarray_output.py
rabernat: you could do something like this

```python
transformations = {
    'k': lambda x: x.k[0, :]
}

# more verbose verion - does exactly the same thing
def get_k_from_model(m):
    return m.k[0, :]
transformations = {
    'k': get_k_from_model
}



# then inside the function
for k, func in transformations.items():
    data[k] = func(m)
```

pyqg/xarray_output.py
rabernat, line 131:
```suggestion
                data = getattr(m, cname, None)
                if data:
                    # do something
```

pyqg/xarray_output.py
hscannell: Wonderful!

rabernat:
Let me know if you're looking for more guidance on how to move forward with this. Once all the comments in my review are resolved, we can merge this PR.

hscannell:
Thanks Ryan! Just adding some tests and I will be done shortly :)

pyqg/xarray_output.py
hscannell: I've combined the loops for vname and cname [here](https://github.com/pyqg/pyqg/pull/230/commits/274fb879275bcd373a6ef230eeddcc1f38664c3d) (see code snippet below). However I'm now essentially filling in the Dataset within the loops and I'm not sure if this is the best way of going about it. 

```python 
def model_to_dataset(m):

ds = xr.Dataset()
    
    # Create list of variables
    variables = {}
    for vname in dim_database:
        if hasattr(m,vname):
            data = getattr(m,vname, None)
            if 'time' in dim_database[vname]:
                variables[vname] = (dim_database[vname],data[np.newaxis,...])
                ds = xr.merge([ds, ds.assign(variables)])
                ds.data_vars[vname].attrs = var_attr_database[vname]
            else:
                variables[vname] = (dim_database[vname],data)
                ds = xr.merge([ds, ds.assign(variables)])
                ds.data_vars[vname].attrs = var_attr_database[vname]

    # Create list of coordinates
    coordinates = {}
    for cname in coord_database:
        if cname in transformations:
            coordinates[cname] = transformations[cname](m)
            ds = xr.merge([ds, ds.assign_coords(coordinates)])
            ds.coords[cname].attrs = coord_attr_database[cname]
        else:
            if hasattr(m,cname):
                data = getattr(m, cname)
                coordinates[cname] = (coord_database[cname],data)
                ds = xr.merge([ds, ds.assign_coords(coordinates)])
                ds.coords[cname].attrs = coord_attr_database[cname

    ds.attrs = ds_attr_database 

    return ds
```

rabernat: I would do it like this

```python
for vname in dim_database:
    # do stuff
    variables[vname] = (dims, data, attrs)
for cname in coord_database:
    # do stuff
    coords[cname] = (dims, data, attrs
ds = xr.Dataset(variables, coords=coords, global_attrs)
```

hscannell:
My test failed because it looks like we need to add xarray as a dependency.
```ModuleNotFoundError: No module named 'xarray'```

File: pyqg/xarray_output.py (hunk 8 above)
Comment (hscannell) [reply pyqg/xarray_output.py]: Made this change, thanks!

File: pyqg/xarray_output.py (hunk 5 above)
Comment (hscannell) [reply pyqg/xarray_output.py]: done.

File: pyqg/tests/test_xarray_output.py [hunk 10]
```diff
@@ -0,0 +1,13 @@
+from __future__ import print_function
+import numpy as np
+import pyqg
+import xarray as xr
+
+def test_xarray():
+    m = pyqg.QGModel(1)
+    ds = m.to_dataset()
+    
+    assert type(ds) == xr.Dataset
```
rabernat, line 10: Can we also assert that certain variables are present. E.g.
```python
expected_vars = ['q', 'phi']  # etc
for v in expected_vars:
    assert v in ds
```

pyqg/tests/test_xarray_output.py
```diff
@@ -0,0 +1,13 @@
...
+
+if __name__ == "__main__":
+    test_xarray()
```
rabernat, line 13: This block is not necessary. The test will run without it.

pyqg/xarray_output.py
```diff
@@ -0,0 +1,138 @@
...
+
+    # Create a dictionary of coordinates
+    coordinates = {}
+    for cname in coord_database:
+        if cname in transformations:
+            coordinates[cname] = (coord_database[cname], transformations[cname](m), coord_attr_database[cname])
+        else:
+            if hasattr(m,cname):
+                data = getattr(m, cname)
+                coordinates[cname] = (coord_database[cname], data, coord_attr_database[cname])
```
rabernat, line 134: This is complete nit picking, but I would refactor this block as follows
```python
for cname in coord_database:
    if hasattr(m, cname):
        if cname in transformations:
            data = transformations[cname](m)
        else:
            data = getattr(m, cname)
        coordinates[cname] = (coord_database[cname], data, coord_attr_database[cname])
```

rabernat (commented):
This looks great! Just a few small suggestions and we are there.

pyqg/tests/test_xarray_output.py
hscannell: OK, thanks. I will remove it.

pyqg/tests/test_xarray_output.py
hscannell: I added this assertion with a few variables I expect to always be present.

Issue #231, opened 2021-03-15
Issue building binder image


jrbourbeau:
I tried to run the binder example linked to in the README but ran into

```
 ModuleNotFoundError: No module named 'Cython'
```

during the `pip install` step when binder attempted to build the corresponding docker image. I've attached the full logs from the binder build below

<details>
<summary>Full build logs:</summary>

```
Waiting for build to start...
Picked Git content provider.
Cloning into '/tmp/repo2dockerkckaoobn'...
HEAD is now at 2937732 Merge pull request #8 from crocha700/binder
Building conda environment for python=3.7Using PythonBuildPack builder
Building conda environment for python=3.7Building conda environment for python=3.7Step 1/49 : FROM buildpack-deps:bionic
 ---> 17c4791bcf21
Step 2/49 : ENV DEBIAN_FRONTEND=noninteractive
 ---> Using cache
 ---> 67552ab561d7
Step 3/49 : RUN apt-get -qq update &&     apt-get -qq install --yes --no-install-recommends locales > /dev/null &&     apt-get -qqpurge &&     apt-get -qq clean &&     rm -rf /var/lib/apt/lists/*
 ---> Using cache
 ---> 197549b860fa
Step 4/49 : RUN echo "en_US.UTF-8 UTF-8" > /etc/locale.gen &&     locale-gen
 ---> Using cache
 ---> 762b9a67ee4f
Step 5/49 : ENV LC_ALL en_US.UTF-8
 ---> Using cache
 ---> f32c81e45755
Step 6/49 : ENV LANG en_US.UTF-8
 ---> Using cache
 ---> fb01e11c213d
Step 7/49 : ENV LANGUAGE en_US.UTF-8
 ---> Using cache
 ---> 99216f5237b0
Step 8/49 : ENV SHELL /bin/bash
 ---> Using cache
 ---> 4010a105965d
Step 9/49 : ARG NB_USER
 ---> Using cache
 ---> 9fd75fb45f4e
Step 10/49 : ARG NB_UID
 ---> Using cache
 ---> 59e03289507d
Step 11/49 : ENV USER ${NB_USER}
 ---> Using cache
 ---> 0aac6f314bc6
Step 12/49 : ENV HOME /home/${NB_USER}
 ---> Using cache
 ---> 1e4970c114d8
[...]

Issue #224, opened 2021-03-09
Create binder for pyqg


rabernat:
Now that there is a conda-forge pyqg package, it would be great to create a binder! It could be based on the examples from the docs.

jamespolly:
Would like to help here but first want to make sure I understand the need. Will the following accomplish our task here?
1. Make PR on `pyqg_demos` repo.
2. Copy over the notebooks and `*.py` dependencies from `pyqg/pyqg_examples` to `pyqg_demos/demos`.

PR #232, opened 2021-03-15
Binder build fixes

jrbourbeau:
It looks like the issue reported over in https://github.com/pyqg/pyqg/issues/231 is due `pyqg` being included in `requirements.txt`. Binder uses [`repo2docker`](https://repo2docker.readthedocs.io/en/latest/) to build docker images based on the contents of a git repository. Since `pyqg` has a `setup.py` file, `repo2docker` will always automatically install `pyqg` so it doesn't need to be included in `requirements.txt`. 

That said, I'm not sure if `requirements.txt` is used elsewhere. Doing a quick grep "requirements.txt" didn't turn up anything, but though it'd be good to double check that removing `pyqg` from `requirements.txt` won't unintentionally break something else. 

Closes https://github.com/pyqg/pyqg/issues/231

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -46,8 +46,8 @@ Links
 .. |landscape| image:: https://landscape.io/github/pyqg/pyqg/master/landscape.svg?style=flat
    :target: https://landscape.io/github/pyqg/pyqg/master
    :alt: Code Health
-.. |binder| image:: http://mybinder.org/badge.svg
-   :target: http://mybinder.org/repo/crocha700/pyqg
+.. |binder| image:: https://mybinder.org/badge_logo.svg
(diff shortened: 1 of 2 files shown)


rabernat:
Thanks for digging into this James! As mentioned in #224, this package is a bit stale and could use quite a bit of overall cleanup.

Most relevant to this PR is the fact that we now have a conda-forge package (which uses the optional but much faster pyfftw) dependency, so going forward we should switch to an `environment.yml` for the binder.

jrbourbeau:
Ah, I see. I missed https://github.com/pyqg/pyqg/issues/224 -- thanks for the additional context

PR #244, opened 2021-05-07
add attributes and copy variables 

hscannell:
This PR addresses #235 and #243.

Issue #243, opened 2021-05-07
variables need to be copied with run with snapshots


hscannell:
For example:
```python
for snapshot in m.run_with_snapshots(tsnapstart=m.t, tsnapint=m.dt):
        model = m.to_dataset()
        model = model.assign(q=(('time','lev','y','x'), m.q.copy()[np.newaxis,...]))
```

PR #245, opened 2021-05-07
Refactor diagnostics for xarray compatibility

### LINKED ISSUE #235: add xarray attributes
Some of the coordinate variables should be attributes. It should look like this:

**Coordinates**
k: 33
l: 64
lev: 2
x: 64
y: 64

**Attributes**
L: float64
W: float64
dt: timedelta64[ns]
filterfac: float64
nk: int64
nl: int64
ntd: int64
nx: int64
ny: int64
nz: int64
pmodes: int64
radii: int64
rek: float64
taveint: timedelta64[ns]
tavestart: timedelta64[ns]
tc: timedelta64[ns]
time: timedelta64[ns]
tmax: timedelta64[ns]
tsnapint: timedelta64[ns]
tsnapstart: timedelta64[ns]
twrite: int64

### LINKED ISSUE #238: refactor / improve diagnostics for xarray compatibility
Our diagnostics allow for specifying units and description

https://github.com/pyqg/pyqg/blob/02e8e713660d6b2043410f2fef6a186a7cb225a6/pyqg/model.py#L611-L626

But these are not populated consistently. 

https://github.com/pyqg/pyqg/blob/02e8e713660d6b2043410f2fef6a186a7cb225a6/pyqg/model.py#L573-L576

We also need to add "dimensions" to the diagnostics metadata. This will give us everything we need to populate the xarray outputs for diagnostics:

https://github.com/pyqg/pyqg/blob/master/pyqg/xarray_output.py

TODO:
- [x] Update existing diagnostics to always provide units and description
- [x] Add dimensions to diagnostic metadata api
- [x] Update existing diagnostics to always provide dims
- [ ] Add a documentation page about diagnostics and how to create a custom diagnostic

hscannell:
The PR closes #238 and closes #235 which adds `dims` and `units` to diagnostics in `model.py`. 

- [ ] Need to check that units are correct

**Looks like there are only 5 core diagnostics (`Ensspec`, `KEspec`, `q`, `EKEdiss`, and `EKE`).** 

Some questions:
(a) why is `q` included as a diagnostic? Isn't it available directly from the model?
(b) where are the other diagnostics defined (e.g., `APEflux`, `APEgen`, `APEgenspec`, `KEflux`, `entspec`)?


rabernat:
> (a) why is `q` included as a diagnostic? Isn't it available directly from the model?

Doesn't make sense to me! I would just remove it.

> (b) where are the other diagnostics defined (e.g., `APEflux`, `APEgen`, `APEgenspec`, `KEflux`, `entspec`)?

Some of the model subclasses define diagnostics that are only relevant to that model subclass. E.g.

https://github.com/pyqg/pyqg/blob/02e8e713660d6b2043410f2fef6a186a7cb225a6/pyqg/qg_model.py#L244-L248

rabernat:
Will have time to look at this tomorrow or Wednesday.

hscannell:
Thanks Ryan! I'd like some advice on how to add these diagnostics to [xarray_output.py](https://github.com/pyqg/pyqg/blob/master/pyqg/xarray_output.py). Here's the list I have...

```python
# dict of diagnostics 
diagnostic_database = {
    'KEspec_modal': (),
    'PEspec_modal': (),
    'KEspec': (),
    'APEspec': (),
    'KEflux': (),
    'KEflux_div': (),
    'APEflux': (),
    'APEflux_div': (),
    'APEgenspec': (),
    'APEgen': (),
    'ENSgenspec': (),
    'ENSflux': (),
    'Ensspec': (),
    'entspec': (),
    'EKEdiss': (),
    'EKE': (),
}
```

How do we grab the dims, description, units from the diagnostics API? Furthermore, how would we deal with diagnostics created by the user on the fly?

hscannell:
I'm envisioning something like this:

```python
diagnostics = {}
for dname in diagnostic_database:
    if dname in m.get_diagnostic:
        data = m.get_diagnostic(dname)
        diagnostics[dname] = (data['dims'], data, {'long_name': data['description'], 'units': data['units']})
```

and then append `diagnostics` to the `variables` dictionary.

pyqg/layered_model.py
```diff
@@ -275,48 +275,75 @@ def _initialize_model_diagnostics(self):
         self.add_diagnostic('entspec',
                 description='barotropic enstrophy spectrum',
                 function= (lambda self:
-                    np.abs((self.Hi[:,np.newaxis,np.newaxis]*self.qh).sum(axis=0))**2/self.H) )
+                    np.abs((self.Hi[:,np.newaxis,np.newaxis]*self.qh).sum(axis=0))**2/self.H),
+                units='meters squared per seconds squared',
```
rabernat, line 279: For units, I recommend we use [UDUNITs conventions](https://www.unidata.ucar.edu/software/udunits/udunits-2.2.28/udunits2.html). They are not very well documented unfortunately. But I think it's something like this

```suggestion
                units='m2 s-2',
```

rabernat (commented):
This is great! Just one comment about your unit syntax.

pyqg/layered_model.py
rabernat: And the same for all units.

rabernat:
> How do we grab the dims, description, units from the diagnostics API?

The model object has a dictionary called `diagnostics` that you should be able to iterate:

```python
for diag_name in m.diagnostics:
    ....
```

pyqg/layered_model.py
hscannell: I took a look at the UDUNITS documentation and they have exponents designated with `^` and the denominator with `/`, so this example would be `m^2/s^2` if I'm not mistaken.

hscannell:
This latest commit 

1. corrects our unit convention to comply with [udunits](https://www.unidata.ucar.edu/software/udunits/udunits-2.2.28/udunits2.html)
2. reformats attributes as a dictionary (#235)
3. adds diagnostic as variables in the xarray dataset using the model object dictionary `diagnostics`
4. calls `.copy()` to the coordinates and variables to update the dataset (#243)

I will close #244 since those changes are rolled into this PR.

hscannell:
I had to amend `test_xarray_output.py` by specifying `tmax`, `twrite`, and `tavestart` for it to pass with the new diagnostic output. Are there any reasons why being this specific in the test is a bad idea?

https://github.com/pyqg/pyqg/blob/d1259cd19b40d586a89781b48d5425678cb77805/pyqg/tests/test_xarray_output.py#L6-L30

pyqg/xarray_output.py
```diff
@@ -126,13 +104,30 @@ def model_to_dataset(m):
...
+        data = transformations[cname](m).copy()
+        coordinates[cname] = (coord_database[cname], data, coord_attr_database[cname])
+
+    # Create a dictionary of global attributes
+    global_attrs = {}
+    for aname in attribute_database:
+        if hasattr(m, aname):
+            data = getattr(m, aname)
+            long_name = attribute_database[aname]
+            global_attrs[long_name] = (data)
```
rabernat, line 116: Hmmm, I'm wondering about this. The problem with using `long_name` as the key here is that you lose the the actual name of the parameter. That would make it very hard to reinitialize a model based on the pickup file.

A different option would be to just use `aname` as the key and discard `long_name`.

hscannell: Thanks for the feedback and that's a good point. I've reverted back to using the parameter key name and made `attribute_database` a list instead of a dictionary. I also changed the test to reflect this and now it passes.

hscannell:
When using this new enhancement, I get an error using `model.to_dataset()` with `model.run_with_snapshots()`. I believe this is because the diagnostics aren't calculated early on and so `model.get_diagnostic()` throws a key error. 

```python
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-7-cda874afc3e3> in <module>
      1 for snapshot in m.run_with_snapshots(tsnapstart=m.t, tsnapint=m.dt):
----> 2     model = m.to_dataset()

~/pyqg/pyqg/model.py in to_dataset(self)
    686         """
    687         from .xarray_output import model_to_dataset
--> 688         return model_to_dataset(self)

~/pyqg/pyqg/xarray_output.py in model_to_dataset(m)
    118     for diag_name in m.diagnostics:
    119         dims = m.diagnostics[diag_name]['dims']
--> 120         data = m.get_diagnostic(diag_name)
    121         if isinstance(data, np.float64):
    122             data = np.array([m.get_diagnostic(diag_name)])

~/pyqg/pyqg/model.py in get_diagnostic(self, dname)
    657 
    658     def get_diagnostic(self, dname):
--> 659         return (self.diagnostics[dname]['value'] /
    660                 self.diagnostics[dname]['count'])
    661 

KeyError: 'value'
```

hscannell:
My workaround is to use a try/except statement when populating the diagnostic variables and it seems to work.
```python
diagnostics = {}
for diag_name in m.diagnostics:
try:
    dims = m.diagnostics[diag_name]['dims']
    data = m.get_diagnostic(diag_name)
    if isinstance(data, np.float64):
        data = np.array([m.get_diagnostic(diag_name)])
    attrs = {'long_name': m.diagnostics[diag_name]['description'], 'units': m.diagnostics[diag_name]['units'],}
    diagnostics[diag_name] = (dims, data, attrs)
except:
    pass
```

pyqg/xarray_output.py
```diff
@@ -126,13 +104,32 @@ def model_to_dataset(m):
...
-            coordinates[cname] = (coord_database[cname], data, coord_attr_database[cname])
+        data = transformations[cname](m).copy()
+        coordinates[cname] = (coord_database[cname], data, coord_attr_database[cname])
+
+    # Create a dictionary of global attributes
+    global_attrs = {}
+    for aname in attribute_database:
+        if hasattr(m, aname):
+            data = getattr(m, aname)
+            global_attrs[aname] = (data)
```
rabernat, line 115:
```suggestion
            global_attrs[f"pyqg:{aname}"] = (data)
```

pyqg/tests/test_xarray_output.py
```diff
@@ -13,4 +21,10 @@ def test_xarray():
     for v in expected_vars:
         assert v in ds
     
-    
+    expected_attrs = ['L', 'W', 'dt', 'filterfac', 'reference'] 
+    for a in expected_attrs:
+        assert a in ds.attrs
```
rabernat, line 26:
```suggestion
        assert f"pyqg:{a}" in ds.attrs
```

rabernat:
We can define a custom error for pyqg to raise when a diagnostic has not yet been filled.

```python
# inside module errors.py

class PQQGError(Exception):
    """Base class for all PYQG exceptions"""
    pass

class DiagnosticNotFilledError(PYQGError):
    """Raise this when trying access a diagnostic that has not been filled"""

# inside model.py:get_diagnostic
if 'value' not in self.diagnostics[dname]:
    raise DiagnosticNotFilledError(f"Diagnostic {dname} has not been filled. Please adjust tavestart, taveint.")

# inside xarray_output.py
from pyqg.errors import DiagnosticNotFilledError
try:
    data = m.get_diagnostic(diag_name)
except DiagnosticNotFilledError:
    pass
```

rabernat:
You might need to add a little bit more code to attach an error message to `DiagnosticNotFilledError`

https://www.programiz.com/python-programming/user-defined-exception

pyqg/tests/test_xarray_output.py
```diff
@@ -4,7 +4,15 @@
...
 def test_xarray():
-    m = pyqg.QGModel(1)
+    # Set parameters & initialize model
+    year = 24*60*60*360.
+    twrite = 1000
+    tavestart = 2*year
+
+    # Initialize QG Model
+    m = pyqg.QGModel(tmax=5*year, twrite=twrite, tavestart=tavestart)
+    m.run()
```
rabernat, line 14: Two comments:
- You don't have to run the model for 5 years to test this feature! Just a few timesteps is enough.
- You should make sure to cover both cases: before _and after_ the diagnostics have been initialized. So you probably need `run_with_snapshots`.

pyqg/layered_model.py
```diff
@@ -275,48 +275,75 @@ def _initialize_model_diagnostics(self):
...
+                dims=('l','k')
+        )
 
         self.add_diagnostic('KEspec_modal',
-                description='modal KE spectra',
+                description='modal kinetic energy spectra',
                 function= (lambda self:
-                    self.wv2*(np.abs(self.phn)**2)/self.M**2 ))
+                    self.wv2*(np.abs(self.phn)**2)/self.M**2 ),
+                units='J',
```
rabernat, line 287: These units are not right. We need to look at the mathematical expression (the function) to understand what the units should be.

- It starts with `phn`, the modal project of `ph`, which should have the same usits as `ph`: `m2 s-1`.
- We square that: -> `m4 s-2`
- Divide by `M**2`: M has no units, just the number of points in the domain (a normalization)
- Multiply by `wv2` (the wavenumber squared; units `m-2`), so we get to `m2 s-2`

In fluids, we always talk about energy _density_, which is technically `J m-3`, but for boussinesq fluids, we always factor out the density, so divide energy by rho (units `kg m^-3`) which brings us to `m2 s-2` (`J = N s-1 = kg m2 s-2`)

rabernat:
Just wanted to chime and and say 👏 ! It looks like you are on the right track here. Let me know when you are ready for another review.

pyqg/errors.py
```diff
@@ -0,0 +1,12 @@
...
+    pass
+
+class DiagnosticNotFilledError(PYQGError):
+    """Raise this when trying access a diagnostic that has not been filled"""
+    def __init__(self, dname):
+        self.dname = dname
+        self.message = f"Diagnostic {dname} has not been filled. Please adjust tavestart and taveint."
+        super().__init__(self.message)
+    
+    pass
```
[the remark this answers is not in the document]
rabernat: You should only specify this message in one place. Right now, from model.py, you are raising this exception as

```python
raise DiagnosticNotFilledError(f"Diagnostic {dname} has not been filled. Please adjust tavestart and taveint.")
```

The string you are passing there will be interpreted as `dname` here. So the message will be

```
"Diagnostic Diagnostic {dname} has not been filled. Please adjust tavestart and taveint has not been filled. Please adjust tavestart and taveint."
```

Can you actually try this code to see what exception is raised? If so, that should help you debug.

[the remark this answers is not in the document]
hscannell: It took me several tries, but I think I have it working now.

pyqg/tests/test_xarray_output.py
hscannell: Using `run_with_snapshots` and testing `m.to_dataset()` before and after diagnostics are calculated. 

https://github.com/hscannell/pyqg/blob/cb1d959589daa7ce8f490ea271637fca3df9390b/pyqg/tests/test_xarray_output.py#L6-L88

pyqg/errors.py
```diff
@@ -0,0 +1,10 @@
+class PYQGError(Exception):
+    """Base class for all PYQG exceptions"""
+    pass
+
+class DiagnosticNotFilledError(PYQGError):
+    """Raise this when trying access a diagnostic that has not been filled"""
+    def __init__(self, dname):
+        self.dname = dname
+        super().__init__(self.dname)
```
rabernat, line 9: There are two ways you could do this. The one you have now is just passing a raw message

```python
class DiagnosticNotFilledError(PYQGError):
    def __init__(self, message):
        super().__init__(message)
```python

OR

```python
class DiagnosticNotFilledError(PYQGError):
    def __init__(self, dname):
        message = f"Diagnostic {dname} has not been filled. Please adjust tavestart and taveint."
        super().__init__(message)
```

rabernat: for the second example, when raising the error, you would just do

```python
dname = "enstrophy_flux"
raise DiagnosticNotFilledError(dname)
# -> "Diagnostic enstrophy_flux has not been filled. Please adjust tavestart and taveint."
```

rabernat:
Just making a note that we are currently waiting for @hscannell to double check the units on the diagnostics.

rabernat:
We can also consider moving forward with blank units for diagnostics we can't figure out and then opening a new issue to keep track of that.

hscannell:
I am updating the documentation examples using xarray and `m.to_dataset()`. For the [layered example](https://github.com/hscannell/pyqg/blob/refactor-diagnostics/docs/examples/layered.ipynb), I'm getting an error for the dimension `lev`.  

```python 
--------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-6-77317b9578bd> in <module>
----> 1 m_ds = m.to_dataset()

~/Documents/projects/pyqg/pyqg/model.py in to_dataset(self)
    689         """
    690         from .xarray_output import model_to_dataset
--> 691         return model_to_dataset(self)

~/Documents/projects/pyqg/pyqg/xarray_output.py in model_to_dataset(m)
    132     variables.update(diagnostics)
    133 
--> 134     ds = xr.Dataset(variables, coords=coordinates, attrs=global_attrs)
    135     ds.attrs['title'] = 'pyqg: Python Quasigeostrophic Model'
    136     ds.attrs['reference'] = 'https://pyqg.readthedocs.io/en/latest/index.html'

~/opt/anaconda3/envs/pyqg_xarray/lib/python3.8/site-packages/xarray/core/dataset.py in __init__(self, data_vars, coords, attrs)
    685             coords = coords.variables
    686 
--> 687         variables, coord_names, dims, indexes, _ = merge_data_and_coords(
    688             data_vars, coords, compat="broadcast_equals"
    689         )

~/opt/anaconda3/envs/pyqg_xarray/lib/python3.8/site-packages/xarray/core/merge.py in merge_data_and_coords(data, coords, compat, join)
    465     explicit_coords = coords.keys()
    466     indexes = dict(_extract_indexes_from_coords(coords))
--> 467     return merge_core(
    468         objects, compat, join, explicit_coords=explicit_coords, indexes=indexes
    469     )

~/opt/anaconda3/envs/pyqg_xarray/lib/python3.8/site-packages/xarray/core/merge.py in merge_core(objects, compat, join, combine_attrs, priority_arg, explicit_coords, indexes, fill_value)
    617     assert_unique_multiindex_level_names(variables)
    618 
--> 619     dims = calculate_dimensions(variables)
    620 
    621     coord_names, noncoord_names = determine_coords(coerced)

~/opt/anaconda3/envs/pyqg_xarray/lib/python3.8/site-packages/xarray/core/dataset.py in calculate_dimensions(variables)
    204                 last_used[dim] = k
    205             elif dims[dim] != size:
--> 206                 raise ValueError(
    207                     "conflicting sizes for dimension %r: "
    208                     "length %s on %r and length %s on %r"

ValueError: conflicting sizes for dimension 'lev': length 2 on 'PEspec_modal' and length 3 on 'q'
``` 

It seems like `q` has three layers, but `PEspec_modal` only has 2. 

This is how `PEspec_modal` is defined in [layered_model.py](https://github.com/hscannell/pyqg/blob/refactor-diagnostics/pyqg/layered_model.py#L291-L297):

 ```python
       self.add_diagnostic('PEspec_modal',
                description='modal potential energy spectra',
                function= (lambda self:
                    self.kdi2[1:,np.newaxis,np.newaxis]*(np.abs(self.phn[1:,:,:])**2)/self.M**2),
                units='',
                dims=('lev','l','k')
        )
```

Maybe we need to specify a special dim for `PEspec_modal` ?

**Comment by rabernat** [comment]
> Maybe we need to specify a special dim for PEspec_modal ?

Yes this sounds right. I think the potential energy is defined at the layer interface, so maybe something like `lev_mid`?. We would eventually want some clarification from the original author.

**Comment by hscannell** [comment]
Another diagnostics with misaligned dimensions is `ENSflux` (the barotropic enstrophy flux). 
It is defined in l[ayered_model.py](https://github.com/hscannell/pyqg/blob/27bc019e651233acd8c543369bd2655f0468338b/pyqg/layered_model.py#L334-L339) as:

```python
self.add_diagnostic('ENSflux',
                 description='barotropic enstrophy flux',
                 function = (lambda self: (-self.Hi[:,np.newaxis,np.newaxis]*
                              (self.qh.conj()*self.Jq).real).sum(axis=0)/self.H),
                units='',
                dims=('x','y')
        )
```

`x` and `y` both have length 64, but `ENSflux` has dimensions (64,33), which looks like (l,k). Are we sure that ENSflux doesn't have spectral dimensions?

rabernat:
> ENSflux doesn't have spectral dimensions?

I am fairly sure that ENSflux **does have spectral dimensions**. It is a flux in spectral space.

asross:
Just want to comment and say that this PR would be very helpful for something I'm working on right now, even without units for the diagnostics!

rabernat:
I agree that we should try to move this forward with the option of blank (`None`) units for the diganostics we can't figure out.

hscannell:
I believe this PR is set to merge. The diagnostic units are left blank. @rabernat do you want me to merge this or would you rather give it one last review?

pyqg/tests/test_xarray_output.py
```diff
@@ -4,13 +4,85 @@
...
+        'lev',
+        'x',
+        'y',
+        'l',
+        'k',
+    ]
     
+    # Initialize QG Model
+    year = 24*60*60*360.
+    m = pyqg.QGModel(tmax=year/2, twrite=10000, tavestart=year/3)
```
rabernat, line 60: Here we are only testing the `QGModel` class. However, there are many other models affected by this change. Ideally we would use pytest [fixture](https://docs.pytest.org/en/6.2.x/fixture.html#back-to-fixtures) to parametrize the test over each different model. That way we could make sure that at least the new function does not raise any errors.

rabernat (changes requested):
> @rabernat do you want me to merge this or would you rather give it one last review?

My one request before merging would be to ensure that the error described in https://github.com/pyqg/pyqg/pull/245#issuecomment-894401714 is actually covered by our test suite. We want to make sure (and have proof, via tests) that every model configuration works with this new feature.

However, if you don't think you have time for that, we can merge as is and accept the risk.

pyqg/tests/test_xarray_output.py
hscannell: Good idea! I've started on this, but my commit will likely fail since I'm not super familiar with pytest fixtures. Please let me know if I'm on the right track.

hscannell: > Fixture "QGModel" called directly. Fixtures are not meant to be called directly,
> but are created automatically when test functions request them as parameters.

How do you parameterize the fixtures in this case?

rabernat: Just saw this question now! I saw your latest commits, and it looks like you're still investigating this.

The trick is to use a [parametrized fixture](https://docs.pytest.org/en/6.2.x/fixture.html#parametrizing-fixtures).

First just write some standalone regular functions that set up the models, e.g. like you already have (but without the decorator

```python
def SQG():
    '''Initialize Surface Quasi-Geostrophic Model'''
    return pyqg.SQGModel(L=2.*np.pi, nx=512, tmax = 26.005, beta = 0., 
                           Nb = 1., H = 1., rek = 0., rd = None, dt = 0.005,
                           taveint=1, twrite=400, ntd=4)

def two_layer():
    # etc
```

Then write a parametrized fixture like this
```python
@fixture(params=[SQG, two_layer, ...])
def all_models(request)
    # actually call the function
    model = request.param()
    return model
```

Then use this as the input to your test
```python
def test_xarray(all_models):
```

🤯

hscannell: 🙏  thank you!

hscannell:
The one thing I noticed when writing these tests is that the diaganostics differ between the models. For example, `LayeredModel` contains `APEflux_div` and not `APEflux` so the assertion fails. I've commented these out so to only check for the diagnostics that are shared amongst all models.

rabernat:
> I've commented these out so to only check for the diagnostics that are shared amongst all models.

That sounds reasonable.

Another approach would be to have a dictionary that allows you to look up this information for each model.

pyqg/tests/test_xarray_output.py
```diff
@@ -2,15 +2,148 @@
...
+            for a in expected_attrs:
+                assert f"pyqg:{a}" in ds.attrs
+
+            for c in expected_coords:
+                assert c in ds.coords
+
+        if snapshot > tsnapint:
+
+            for v in list(ds.keys()): 
+                assert v in expected_vars + expected_diags
```
hscannell, line 143: Instead I checked that all the dataset keys are in `expected_vars` and `expected_diags`. This catches all models.

asross:
Thanks for merging this!!

PR #254, opened 2021-09-14
Support xarray concatenation of simulation snapshots across time

asross:
This pull request slightly tweaks the handling of time to support the following use-case:

```python
pyqg_outputs_over_time = xr.concat(
    [snapshot.to_dataset() for snapshot in pyqg.QGModel().run_with_snapshots()],
    dim="time"
)
```

The main way it does this is by adding a time dimension to diagnostics (even though they are time-averaged, but averages can still evolve) and updating the time coordinate of the resulting dataset to the current time of the model.

### DIFF
--- a/pyqg/tests/test_xarray_output.py
+++ b/pyqg/tests/test_xarray_output.py
@@ -142,6 +142,10 @@ def test_xarray(all_models):
             for v in list(ds.keys()): 
                 assert v in expected_vars + expected_diags
 
+            for v in expected_diags:
+                if v in ds:
+                    assert 'time' in ds[v].coords
+
             for a in expected_attrs:
                 assert f"pyqg:{a}" in ds.attrs
 
--- a/pyqg/tests/test_xarray_output.py
+++ b/pyqg/tests/test_xarray_output.py
@@ -155,7 +155,6 @@ def test_xarray(all_models):
             for c in expected_coords:
                 assert c in ds.coords
 
-    if len(datasets):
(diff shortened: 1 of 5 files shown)


pyqg/tests/test_xarray_output.py
```diff
@@ -142,8 +148,17 @@ def test_xarray(all_models):
...
+                if v in ds:
+                    assert 'time' in ds[v].coords
+
             for a in expected_attrs:
                 assert f"pyqg:{a}" in ds.attrs
 
             for c in expected_coords:
                 assert c in ds.coords
+
+    if len(datasets):
```
asross, line 161: I had to add this `if` because some of the models in `all_models` actually never get any snapshots given their current settings... this might be an issue with the existing tests!

rabernat:
Thanks a lot for this contribution @asross!

I'll ask @hscannell to review this as she is the expert on the xarray output feature.

pyqg/tests/test_xarray_output.py
hscannell: Thanks @asross for point this out! I've adjusted the model parameters so that they all get snapshots.

hscannell (commented):
Looks great!

pyqg/tests/test_xarray_output.py
asross: Awesome!

asross:
> Thanks so much for this contribution!
> 
> i just have one question about the discussion below...

Done!

Issue #255, opened 2021-09-15
Update packaging


rabernat:
pyqg packaging and release procedure is very old. Ideally we should
- [x] drop python 2.7 support
- [x] move to setuptools_scm-style packaging
- [x] create a GitHub workflow for releasing

hugovk:
Please see PR https://github.com/pyqg/pyqg/pull/258 to drop Python 2.7.

rabernat:
Closed by #258

hugovk:
https://github.com/pyqg/pyqg/pull/258 only addressed the first checkbox, not the other two.

PR #258, opened 2021-09-16
Drop support for EOL Python 2.7

hugovk:
For https://github.com/pyqg/pyqg/issues/255.

Whilst we're here, let's also test Python 3.9.


docs/conf.py
```diff
@@ -83,8 +82,8 @@ def __getattr__(cls, name):
 master_doc = 'index'
 
 # General information about the project.
-project = u'pyqg'
-copyright = u'2015, PyQG team'
+project = 'pyqg'
+copyright = '2015, PyQG team'
```
rabernat, line 86:
```suggestion
copyright = '2015-2021, PyQG team'
```

rabernat (commented):
Thank you so much for the contribution! This sort of maintenance is exactly what the project is sorely missing right now. 😄

PR #260, opened 2021-10-04
Add more verbose ImportError when Xarray not found

### LINKED ISSUE #259: xarray
Apparently conda will install the latest version of pyqg without checking for xarray.  I'm not an expert on conda packages, but shouldn't it be possible to specify dependency on xarray and make sure conda will install it?

rabernat:
Fixes #259

We should also mention this in the documentation.

### DIFF
--- a/pyqg/xarray_output.py
+++ b/pyqg/xarray_output.py
@@ -1,5 +1,13 @@
 import numpy as np
-import xarray as xr
+try:
+    import xarray as xr
+except ImportError:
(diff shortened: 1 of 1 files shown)


rabernat:
We also need an update to the doc examples explaining that Xarray is an optional dependency. That won't be as quick, since it requires editing notebooks, which I can't do from GitHub. @hscannell (who added the xarray feature) is on vacation.

PR #263, opened 2021-10-05
SQG inversion relation correction

### LINKED ISSUE #178: Inversion relation for SQG and barotropic deformation radius
Hi all,

I just started playing around with pyqg (love it!) and think I've stumbled upon two issues:

1) The barotropic deformation radius (self.radii[0]) is computed as self.g * self.H / f, and I think you meant to square root the numerator.

2) I think there's a error in the inversion relationship between buoyancy (b) and the streamfunction (psi) in the SQG model. If I'm not mistaken, it implies that hat(psi) = N/K * hat(b), but I think it should be 1/(NK) * hat(b). For the example scenario you presented from the Held et al (1995) paper, Nb was set to 1 so it would have had no effect on the solution.

Thanks!

SalahKouhen:
I changed the inversion relation in the SQG model, to see why it must be this (the same as what is in the documentation) see my addition to the docs. 

fix #178

### DIFF
--- a/docs/equations/notation_sqg_model.rst
+++ b/docs/equations/notation_sqg_model.rst
@@ -40,7 +40,7 @@ and
 
    \psi = 0,  \qquad \text{at} \qquad z \rightarrow -\infty\,.
 
-The solutions to the elliptic problem above, in horizontal Fourier
+The solutions to the elliptic problem above*, in horizontal Fourier
 space, gives the inversion relationship between surface buoyancy and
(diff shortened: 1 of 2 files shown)


SalahKouhen:
I also removed rd and rek from the arguments (they don't do anything and so are quite misleading!) and added the Coriolis parameter. I could have used rd instead but this is not how the equations are usually presented. The example for SQG has to be changed.

PR #264, opened 2021-10-06
SQG inversion relation 

### LINKED ISSUE #178: Inversion relation for SQG and barotropic deformation radius

SalahKouhen:
fix #178 

I changed the inversion relation in the SQG model, to see why it must be this (the same as what is in the documentation but not the ipython example or comments) see my addition to the docs. 
I also removed rd and rek from the arguments (they don't do anything and so are quite misleading for anyone trying to quickly get started) and added the Coriolis parameter. I could have used rd instead but this is not how the equations are usually presented and figuring out what H should be would be tricky. 
The examples for SQG have been changed to reflect my changes.

### DIFF
--- a/docs/equations/notation_sqg_model.rst
+++ b/docs/equations/notation_sqg_model.rst
@@ -68,7 +68,7 @@ Taking the Fourier transform in the x and y directions with :math:`\kappa^2 = k^
 .. math::
 
 
-    \frac{f_0^2}{N^2}\frac{\partial }{\partial z}\left(\frac{\partial \hat \psi}{\partial z}\right) = \kappa^2 \nabla^2\psi\,,
+    \frac{f_0^2}{N^2}\frac{\partial }{\partial z}\left(\frac{\partial \hat \psi}{\partial z}\right) = \kappa^2 \psi\,,
     
 which has solution
 
@@ -78,5 +78,11 @@ which has solution
    \hat \psi = Ae^{\frac{\kappa N}{f_0}z} + Be^{-\frac{\kappa N}{f_0}z},.
    
 
-Our decay at negative infinity immediately tells us that :math:`B = 0`. Differentiating with respect to :math:`z` and evaluating at the surface tells us :math:`A = f_0 \hat b / \kappa N` so that we have the result. 
+Our decay at negative infinity immediately tells us that :math:`B = 0`. Differentiating with respect to :math:`z` and evaluating at the surface tells us :math:`A = f_0 \hat b / \kappa N` so that we have:
 
+.. math::
+
+
+   \hat \psi = \frac{f_0 \hat b}{\kappa N }e^{\frac{\kappa N}{f_0}z},.
+
+Evaluating at :math:`z = 0` gives the inversion relation given above. 
--- a/docs/equations/notation_sqg_model.rst
+++ b/docs/equations/notation_sqg_model.rst
@@ -60,7 +60,7 @@ model.
 .. math::
 
 
-    \frac{\partial }{\partial z}\left(\frac{f_0^2}{N^2}\frac{\partial \psi}{\partial z}\right) + \nabla^2\psi = 0\,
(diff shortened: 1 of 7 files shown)


SalahKouhen:
(Note that it is failing the tests as I removed rd as an argument which is what I intend.)

rabernat:
Thanks a lot for this PR @SalahKouhen! We really appreciate your attention to these details.

Could you go ahead and change the test that are failing as well?

I am very busy with teaching right now and am not sure I will have time to review this soon. Perhaps @MFJansen could have a look?

rabernat:
> Not sure why I have a very slight difference to the reference statistic: 7847.664033 compared to 7847.516961

These small differences in the verification tests are not important. Due to imprecise floating-point arithmetic, we don't expect every model run to be identical on different machines. Feel free to turn down the tolerance in order to make the test pass.

francispoulin:
I am happy to take a look.  Can someone remind me how to view the docs for this PR?  It would be easier for me to read that then read the latex.  Thanks!

francispoulin:
> Yeah, should 100% be a z not a t in the partial derivative. Just a typo.
> 
> For the equations as they are presented in the docs (nowhere in the docs or comments does it say we are putting it in a non-dimensional form) this is the correct inversion relation.

Thanks for fixing the typo, and that does make more sense, and I agree with the solution.

francispoulin:
> @francispoulin f_0/N is non-dim so if b is non-dim doesn't this mean the new form is non-dim? Trying to get my head around this haha. I'm mostly looking at the held paper from 1995 and a review from Lapeyre in 2017
> 
> Edit: Not as important as the code but where are you seeing mode instead of model?

I agree that `f_0/N` is non-dimensional but the variables you are using, `f_0, N`, are dimensional.  Looking in Lapeyre (2017), he uses a dimensional formulation of the model at the beginning. A non-dimensional formulation would mean that `z` and `\kappa` would have dimensions of `H` and `L`, say, respectively.  

Non-dimensionalizing your second last equation would become something like

\partial_{zz} \hat \psi = Bu^2 \nabla^2 \hat \psi,

and then the solution would be

\psi = A e^{Bu z} + B e^{-Bu z}. 

Unfortunately, the code that I put together did not do this but I think that would have been clearner.  Sorry for the messy code.  

Lapeyre (2017) changes to non-dimensional equations after equation (23), and he does this by introducing a transformed length z N/f_0.  His equation (24)  does not match your equation for two reasons.  First, he has imposed his equation (21) to neglect the term that grows moving away from the surface.  Second, he chose the coefficent bases on the surface condition, which is your equation at z=0. 

Do you think that starting off with a non-dimensional form using a `Bu` would be a good idea? 

I think that putting your equations more in alignment with his paper would be very nice, as that is a very good reference that people use a lot.

francispoulin:
Following @MFJansen , for a dimensional formulation, I would suggest the following:

- Fixing typo in second last equation, the right hand side shoulde be - \kappa^2, no \nabla^2
- I would suggest writing out the final form of the equation with B = 0 and A as it's defined, would be nice to see.

SalahKouhen:
> You will need the [documentation requirements](https://github.com/pyqg/pyqg/blob/master/docs/requirements.txt) installed in your environment
> 

I had some trouble with this, I ended up needing to install nbsphinx and sphinx_rtd_theme separately.  Very similar to the issue here https://github.com/mathLab/PyGeM/issues/94 .

I have made a few tweaks, does this look good to get merged?

rabernat:
Sorry for the very late response to this PR. We are trying to revive development on pyqg. Thanks for your patience!

I just reviewed all the changes and I think this PR is ready to merge!

Issue #237, opened 2021-05-06
provide hooks for custom forcing functions


rabernat:
@arthurBarthe is making predictions for extra momentum flux tendencies using an offline data-driven model. We need to figure out a good API to allow users to plug in custom forcing / predictions.

Related to #85  #134

PR #266, opened 2021-10-08
Add support for online parameterizations / custom forcing functions

asross:
We're starting to use pyqg to collaborate on different methods for implementing subgrid forcing parameterizations, which we currently have implemented in two different ways (both velocity and potential vorticity directly). This pull request would add functionality to support those use-cases via `q_parameterization` and `uv_parameterization` hooks, which take the model as input and return either updates to q or u/v tendencies (in real space), which we convert to spectral space and apply to `m.dqhdt`.

Clearly, this is a complicated feature, and needs tests / probably needs API design improvements. However, I wanted to make this pull request to start the conversation.

This builds on work done by @arthurBarthe in https://github.com/arthurBarthe/pyqg/tree/adv_param and would potentially resolve https://github.com/pyqg/pyqg/issues/237.


pyqg/kernel.pyx
```diff
@@ -55,6 +55,15 @@ cdef class PseudoSpectralKernel:
...
     cdef DTYPE_com_t [:, :, :] dqhdt_pp
 
+    # subgrid velocity parameterizations
+    cdef DTYPE_real_t [:, :, :] du
+    cdef DTYPE_real_t [:, :, :] dv
+    cdef DTYPE_com_t [:, :, :] duh
+    cdef DTYPE_com_t [:, :, :] dvh
+    # subgrid potential vorticity parameterizations
+    cdef DTYPE_real_t [:, :, :] dq
+    cdef DTYPE_com_t [:, :, :] dqh
```
rabernat, line 65: Since there is no state about the parameterization that has to be persisted, I am curious why we need to add all of this stuff to the base model class.

Maybe the goal is to be able to include this stuff in diagnostics?

pyqg/kernel.pyx
```diff
@@ -347,6 +388,53 @@ cdef class PseudoSpectralKernel:
...
+        cdef Py_ssize_t k, j, i
+        du, dv = self.uv_parameterization(self)
+        cdef DTYPE_real_t [:, :, :] du_view = du
+        cdef DTYPE_real_t [:, :, :] dv_view = dv
+        # convert to cython memory view
+        self.du[:, :, :] = du_view
+        self.dv[:, :, :] = dv_view
+        # convert to spectral space
+        self.fft_du_to_duh()
+        self.fft_dv_to_dvh()
```
rabernat, line 405: Could we not just use function-local variables here?

rabernat (commented):
Hi @asross - thanks so much for getting this important feature started! Your implementation looks pretty good to me.

Before moving forward, I would say it's pretty important to get a test for such a big new feature. We need the test to "prove" that this feature works and to verify that we don't break it with future changes.

The tests doesn't have to be a "real" parameterization in all its complexity. Just something that triggers these parts of the code to run. Let me know if you need some advice about where. to put the test.

pyqg/kernel.pyx
asross: `dqh`, `duh`, and `dvh` are being used in diagnostics [here](https://github.com/pyqg/pyqg/blob/7d50c93fdfc23997cfed2d691b102344f3b7f1ff/pyqg/model.py#L623-L665). We could omit their real counterparts, however, if we also follow your suggestion below to use function-local variables. I'm not certain about whether that would be less efficient, though (e.g. it seems like there might be performance benefits to initializing the FFT plans and memory for the arrays once at the top, rather than reinitializing / reallocating everything every timestep).

pyqg/kernel.pyx
asross: See comment above -- I don't have a great sense of the C performance considerations here.

asross:
> Before moving forward, I would say it's pretty important to get a test for such a big new feature. We need the test to "prove" that this feature works and to verify that we don't break it with future changes.
>
> The tests doesn't have to be a "real" parameterization in all its complexity. Just something that triggers these parts of the code to run. Let me know if you need some advice about where. to put the test.

Totally -- I agree this needs enough tests to fully cover the contribution. Just wanted to confirm first that you think the API for adding these kinds of parameterizations is sufficiently flexible / intuitive!

rabernat:
> Just wanted to confirm first that you think the API for adding these kinds of parameterizations is sufficiently flexible / intuitive!

Yes it looks good to me! And I understand why you are doing it this way (adding new variables to the model).

I'm a bit concerned about the memory footprint. This will add some memory overhead to all models, regardless of whether they use the feature or not. But I suppose that is unavoidable.

asross:
> I'm a bit concerned about the memory footprint. This will add some memory overhead to all models, regardless of whether they use the feature or not. But I suppose that is unavoidable.

Hmm... I think we could avoid some overhead by modifying the signature of `PseudoSpectralKernel.__init__` to include boolean indicators of whether the model is parameterized (passing them in [here](https://github.com/pyqg/pyqg/blob/7d50c93fdfc23997cfed2d691b102344f3b7f1ff/pyqg/model.py#L175)). Then we could avoid some unnecessary `empty_real` and `pyfftw.FFTW` calls by making them conditional. We still have extra `cdef` calls but maybe those don't add any memory if they're just pointers? Not exactly sure how Cython works :)

asross:
@rabernat tests are added, and I also made allocation of some C arrays conditional on whether a parameterization is provided. Note that this does prevent users from assigning a parameterization after initializing a model, but I don't think that's a huge issue.

asross:
Just checking in about this -- would be useful to get this merged in soon!

rabernat:
Hi @asross -- sorry for the very slow review process here! 😓 

I just looked over your PR. It looks good. I have some ideas about how to refactor things to possibly optimize performance, but that can come later.

The one remaining thing that this PR needs is **documentation**. I apologize for not suggesting this sooner--you could have been working on that over the past two weeks. 

At bare minimum, we need to get these new attributes into the `Model` docstring here - https://pyqg.readthedocs.io/en/latest/api.html#module-pyqg. Ideally we would also have some narrative documentation--perhaps a new notebook in [examples](https://github.com/pyqg/pyqg/tree/master/docs/examples)?

I promise this will be the last iteration. Thanks for your patience.

asross:
No worries! I added more docstrings and an example notebook, let me know if those look ok!

Also, I realized when implementing this parameterization that I might be doing an unnecessary IFFT at the end (now that I’m taking derivatives in spectral space), because the kernel will immediately take the FFT again. To fix that, we’d need to change the interface to allow the user to specify whether the correction terms are in real or spectral space.

I’m not sure if fixing that is worth holding up the pull request, though, especially since this is just used for research purposes — so maybe we can defer that to a follow up where we add that option along with your suggested performance improvements.

rabernat:
Great! 

You will need to add your new nb example to the sphinx toctree here in order for it to actually show up in the docs: https://github.com/pyqg/pyqg/blob/master/docs/examples.rst

Hopefully that will "just work". Are you able to build the docs locally?

asross:
Ah, good catch! It's added to the TOCs and everything seems to render properly locally -- though I notice that some of the other examples include a `{notebook}_files` directory with the images. Let me know if that's necessary.

rabernat:
Fantastic! 🚀🚀🚀 Thanks so much for this great PR. 

> though I notice that some of the other examples include a `{notebook}_files` directory with the images.

I think this is a relic of a much older (pre nbsphinx) doc configuration. This package has been around for a while! A PR to clean up those files (I think they are unnecessary would be most welcome).

PR #270, opened 2021-10-15
Add helper method for running a simulation with grids of particles

asross:
While using particle classes to help visualize the results of pyqg simulations, I ended up writing some helper methods for running simulations with particles together, that seemed like they might be a useful feature for the main library. Here's an initial pass at how this might be incorporated, let me know what you think!


rabernat:
@hscannell maybe you could have a look at this since you are doing particles with pyqg?

asross:
Let me know if this looks good to merge @hscannell!

I also added a [quick animated demo](https://pyqg--270.org.readthedocs.build/en/270/examples/particles.html) of this feature to `docs/examples`.

pyqg/model.py
```diff
@@ -233,6 +237,35 @@ def __init__(
         self._initialize_inversion_matrix()
         self._initialize_diagnostics(diagnostics_list)
 
+    def add_particles(self, initial_coords=None, order=1):
```
rabernat, line 240:
```suggestion
    def add_particles(self, density=0.125, initial_coords=None, order=1):
```

pyqg/model.py
```diff
@@ -233,6 +237,35 @@ def __init__(
...
+        """Add particles to the simulation whose positions will be updated at
+        each timestep.
+
+        Parameters
+        ----------
+
```
rabernat, line 246:
```suggestion
        density : float
            How many particles to put per grid cell. Mutually exclusive with ``initial_coords``.
```

pyqg/model.py
```diff
@@ -233,6 +237,35 @@ def __init__(
...
+        initial_coords : tuple
+            An optional tuple of x,y coordinates for the initial positions of
+            particles at each fluid layer.
+        order : int
+            Order of interpolation.
+        """
+
+        if initial_coords is None:
+            n = self.nx // 8
```
rabernat, line 255:
```suggestion
            n = int(self.nx * density)
```

or maybe something even better (non-discretized, using linspace)

rabernat (commented):
This looks awesome. One minor API suggestion which can also be ignored for now.

Issue #272, opened 2021-10-19
`calc_ispec` error at small wavenumbers / large lengthscales


asross:
I'm seeing strange-looking "elbow" artifacts at low wavenumbers when looking at isotropic spectra calculated using the [`calc_ispec` helper](https://github.com/pyqg/pyqg/blob/master/pyqg/diagnostic_tools.py#L77).

I believe this is because the minimum wavenumber (`0.5 * sqrt(Δk^2 + Δl^2)`) is actually too small -- causing the radius of the length scale to go past the edges of the domain. From a quick check, it seems like the elbow occurs at `1.5 * sqrt(Δk^2 + Δl^2)` or two thirds of the maximum radius. However, setting the minimum wavenumber to that value seems to resolve the issue.

![image](https://user-images.githubusercontent.com/1022564/137980114-575e2960-858e-421f-a562-e11ddbbab9b6.png)

Would it make sense to update [this line](https://github.com/pyqg/pyqg/blob/master/pyqg/diagnostic_tools.py#L77) from `dkr/2.` to `1.5*dkr`?

asross:
Closing in favor of https://github.com/pyqg/pyqg/issues/275.

PR #273, opened 2021-10-26
Increase minimum wavenumber in calc_ispec to avoid large-lengthscale artifacts

asross:
See https://github.com/pyqg/pyqg/issues/272.

### DIFF
--- a/pyqg/diagnostic_tools.py
+++ b/pyqg/diagnostic_tools.py
@@ -74,7 +74,7 @@ def calc_ispec(model, ph):
 
     # create radial wavenumber
     dkr = np.sqrt(model.dk**2 + model.dl**2)
-    kr =  np.arange(dkr/2.,kmax+dkr,dkr)
+    kr =  np.arange(1.5*dkr,kmax+dkr,dkr)
     phr = np.zeros(kr.size)
 
     for i in range(kr.size):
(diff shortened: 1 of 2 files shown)


rabernat:
Hi @asross - thanks so much for this!

It looks this area of the code is not currently covered by tests 😬. Would you be willing to perhaps add a quick test that covers the `calc_ispec` function? That way we could have a lot more confidence going forward that this feature will continue to work.

You could either add it to https://github.com/pyqg/pyqg/blob/master/pyqg/tests/test_reference_solns.py or else create a new test module (e.g. `test_diagnostic_tools.py`).

If you don't have time for this, we can still merge it as-is.

asross:
Yep, will do -- it might take a week or two, but I agree it's important to test this area. Hopefully I can even write a test with a function whose isotropic spectrum has ground-truth values, and demonstrate that this minimum wavenumber is necessary to accurately compute it.

rabernat:
Sounds good. I would say that _any_ test is better than no test. There are two separate tests being discussed:
- Whether the function actually works and does what it says it will do (my hope for this PR)
- Whether the model simulates and expected spectrum (what you proposed above) -- this is more of a "verification test" for the model than a test of the new feature itself (and is therefore a lower priority for me)

asross:
Added a basic test that the isotropic spectrum peaks in the right place for a periodic function of known frequency!

PR #274, opened 2021-11-05
Add real streamfunction and real/spectral dqdts to xarray

asross:
I need to save the real-space values of the streamfunction and the PV tendencies locally, so I had written a wrapper around `pyqg.Model#to_dataset` to include them. However, I'm making a pull request to add them directly within the model, since they might be generally useful to include (at least for the streamfunction, the only variable that `to_dataset` includes in spectral but not real space).


rabernat:
This is great! Thanks for the contribution.

Issue #277, opened 2022-02-28
Update version (+ on pip / Anaconda)


asross:
I've gotten some requests to have the newest version of pyqg that supports parameterizations installable via pip / conda. Any chance we could tag and upload a new version?

rabernat:
Hi @asross! Yes this is a good idea. Unfortunately it is not as easy as it should be - see #255.

If you would like to take the lead on the release, please be my guest. I have added you to the maintainers group @pyqg/pyqg_devs.

asross:
Awesome. I should warn that I haven't done too much package publishing in Python (mostly Ruby/JS), so migrating from versioner to setuptools_scm might be a bit tricky -- especially in a way that doesn't break cython or FFTs. So if there is a some kind of existing process for publishing a new package without that overhaul, that would be great. But I'll look into it if not!

rabernat:
For now it is possible to do the release manually. Instructions are here: https://pyqg.readthedocs.io/en/latest/development.html#release-procedure

asross:
Sorry for the slow progress on this -- the first thing I noticed was that the release instructions haven't been updated since versioneer was added, so the way to increment the version number now seems to be to just add the new git tag (`v0.5.0`) to the appropriate commit (and versioneer appears to be able to automatically infer the version number from the tag?).

The second thing I noticed, however, was that `master` currently contains some of my changes to `calc_ispec` (merged after the parameterization PR merge) which, per https://github.com/pyqg/pyqg/issues/275, now seem incorrect. So I was thinking of trying to resolve those first before incrementing the version. However, figuring out the right normalization factor has proved kind of tricky.

My pyqg plan now is to instead:
- assign the 0.5.0 tag to https://github.com/pyqg/pyqg/tree/00990ba51c85c9c2ad3b066ef166154c27ca70bf (when the parameterization PR was merged)
- run through the rest of the release process (i.e. pypi sdist, assuming you adding me to the maintainers group on Github gave me permission to update pypi)
- create a separate PR to update the release instructions and whats-new.rst
- hopefully move on to figuring out calc_ispec normalization and getting that merged (then updating to 0.5.1)
- hopefully move on from there to creating some additional PRs with diagnostics refactors/units and new features :)

rabernat:
Sounds like a great plan @asross! Let me know if you get blocked on anything.

I just added you as an owner on pypi for pyqg.

asross:
[...]

Issue #216, opened 2019-11-27
module 'versioneer' has no attribute ‘get_version’


YONGQUAN-QU:
I'm trying to upgrade pyqg to 0.3.0 after installing versioneer. 
It seems like there is an error

 "Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-install-hmbw2nb3/pyqg/setup.py", line 143, in <module>
        version=versioneer.get_version(),
    AttributeError: module 'versioneer' has no attribute 'get_version'
    check_for_openmp() result:  0"

I wondered how to fix the problem. Thank you very much

navidcy:
+1 -- same problem here.

Btw, why one needs to install `versioneer`package independently? Can't `versioneer` be included in pyqg's dependencies?

uegajde:
i find the reason.
somehow the file on pypi lacks the versioneer.py
https://github.com/pyqg/pyqg/blob/master/versioneer.py

to install:
1. download pyqg-0.3.0.tar.gz  from pypi
2. uncompress it
3. cd pyqg-0.3.0
3. copy the versioneer.py from https://github.com/pyqg/pyqg/blob/master/versioneer.py
4. install with "pip install ."

or just clone the repo and then install with "pip install ."

rabernat:
Sorry for this problem. It is popping up everywhere, for example, see https://github.com/xgcm/xrft/issues/108 for the same problem in a different package. I think it is related to a change in python packaging that I don't fully understand.

illtellyoulater:
I've encountered the same problem. I'm attaching the full output in case you need it for debugging:

```
INFO: pip is looking at multiple versions of opt-einsum to determine which version is compatible with other requirements. This could take a while.
  Using cached opt_einsum-2.1.3.tar.gz (49 kB)
  Using cached opt_einsum-2.1.2.tar.gz (49 kB)
  Using cached opt_einsum-2.1.1.tar.gz (33 kB)
    ERROR: Command errored out with exit status 1:
[...]

PR #278, opened 2022-03-05
use setuptools_scm, fix packaging bugs, and update docs

asross:
This should resolve https://github.com/pyqg/pyqg/issues/277, https://github.com/pyqg/pyqg/issues/216, and https://github.com/pyqg/pyqg/pull/276, and hopefully make it so downloading/uploading pyqg to/from pypi just works (though there's a potential issue with pyfftw vs. np.fft which I'll detail below). I've tested it using testpypi but I'm holding off deploying to the main pypi repository until we've gotten this merged and updated whats-new.rst (PR to follow).

~~I also ~~suspect I might~~ definitely have an issue building the docs, but ~~we'll see what happens when this PR triggers a new build~~ I'll fix that soon.~~ I think I've resolved this issue by just upgrading all of the doc dependencies.

One other potential issue is that we lose the flexibility we previously had around `pyfftw` vs. `np.fftw`. In order for plain old `pip install pyqg` to support `pyfftw` (which I think we should want as the default), we need to specify it in `pyproject.toml`. But, this means users have to download it to install or run pyqg (unless they want to install pyqg by cloning the repository and doing a local install), which makes it effectively a requirement (that also required updating CI to only run tests using it).

Another option is to omit `pyfftw` from `pyproject.toml` but tell `pyfftw` users to run `pip install pyqg --no-build-isolation`. That feels a bit wonky though.


rabernat (commented):
Andrew thanks so much for working on this. I read through this issue:
- https://github.com/pypa/pip/issues/6144#issuecomment-455266791

pretty carefully and concluded that what we want--optional build dependencies--is not supported by pyproject.toml.

TBH this seems like a huge rabbit hole of complexity.

So for now, I am fine with simply requiring pyfftw as a hard dependency. Going forward, maybe we can find a way to make it optional again.

rabernat:
After merging this, I just tried pip installing from git

```
% pip install git+https://github.com/pyqg/pyqg.git
Collecting git+https://github.com/pyqg/pyqg.git
  Cloning https://github.com/pyqg/pyqg.git to /private/var/folders/n8/63q49ms55wxcj_gfbtykwp5r0000gn/T/pip-req-build-l_3uvzg1
  Running command git clone --filter=blob:none --quiet https://github.com/pyqg/pyqg.git /private/var/folders/n8/63q49ms55wxcj_gfbtykwp5r0000gn/T/pip-req-build-l_3uvzg1
  Resolved https://github.com/pyqg/pyqg.git to commit 3b6b15cb282df9924a4be9ac2dd939c17290e8c0
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Collecting cython
  Using cached Cython-0.29.28-py2.py3-none-any.whl (983 kB)
Collecting numpy
  Using cached numpy-1.22.3-cp39-cp39-macosx_10_14_x86_64.whl (17.6 MB)
Building wheels for collected packages: pyqg
  Building wheel for pyqg (pyproject.toml) ... done
  Created wheel for pyqg: filename=pyqg-0.5.1.dev3+g3b6b15c-cp39-cp39-macosx_10_9_x86_64.whl size=161316 sha256=3b00d5b6d42c67864012d5be550418fd4aac13133374f2199e8081e45f7b07d4
  Stored in directory: /private/var/folders/n8/63q49ms55wxcj_gfbtykwp5r0000gn/T/pip-ephem-wheel-cache-kz7s0lvn/wheels/f3/13/1b/57f93b097bcabd35cfb3c25c9129a83588b3d19148c74e4c4f
Successfully built pyqg
Installing collected packages: numpy, cython, pyqg
Successfully installed cython-0.29.28 numpy-1.22.3 pyqg-0.5.1.dev3+g3b6b15c
```

I am puzzled by the fact that pyfftw was not installed.

rabernat:
Ah ok so you're saying it got the dependencies from here:

https://github.com/pyqg/pyqg/blob/3b6b15cb282df9924a4be9ac2dd939c17290e8c0/setup.py#L60-L64

It feels a little redundant to have to specify both. I guess this is why people hate python packaging? 🙃

asross:
> Ah ok so you're saying it got the dependencies from here:
> 
> https://github.com/pyqg/pyqg/blob/3b6b15cb282df9924a4be9ac2dd939c17290e8c0/setup.py#L60-L64
> 
> It feels a little redundant to have to specify both. I guess this is why people hate python packaging? 🙃

A plausible hypothesis 🙃

asross:
`pip install git+https://github.com/pyqg/pyqg.git@more-setup-debugging` should hopefully work now?

PR #280, opened 2022-03-23
Try to support more pip install options

asross:
See comments at the end of https://github.com/pyqg/pyqg/pull/278


rabernat:
I just tested this with ` pip install git+https://github.com/pyqg/pyqg.git@more-setup-debugging` and it worked. (Can successfully import pyqg). So I think we are good here.

PR #281, opened 2022-03-23
Update whats-new.rst

asross:


asross:
Planning on merging this directly once CI completes, since there are no substantive changes.

asross:
Looks like one test is failing, but it looks like a random issue with an array not being quite in tolerance. Will fix in a later commit.

PR #284, opened 2022-03-25
Normalizing & unitizing diagnostics

asross:
This PR will add missing units to diagnostics, and also normalize them (e.g. by `M**2`) such that diagnostics from models with different resolutions but the same domain size are directly comparable.

TODOs:
- [x] add units to diagnostics from other models
- [x] ~~write tests~~ sort of -- see [this notebook](https://github.com/pyqg/pyqg/blob/01b4bb2cd811e04b2868782066eec811a499986c/examples/diagnostic_normalization.ipynb)


rabernat:
Fantastic to see this happening! Thanks so much @asross!

asross:
Alright, this is now ready for review (pending fixing the tests). I _think_ I've gotten all of the units correct, and importantly, I've verified that reducing resolution, while reducing diagnostic extremes a bit, [doesn't totally change their order of magnitude](https://github.com/pyqg/pyqg/blob/01b4bb2cd811e04b2868782066eec811a499986c/examples/diagnostic_normalization.ipynb)--which I think suggests that all quantities are being normalized properly.

It would be good to translate that into a proper test, but if I adapted my verification notebook into one, it would take too long to reasonably run. One option is to save the final states of the models as fixtures, load them in the test, and then run the same tests on instantaneous diagnostics, but it feels a bit kludgy. Anyone have thoughts?

asross:
I'll plan to merge this soon unless there are objections or suggestions -- the normalization changes permit automated comparisons of diagnostics from simulations at different resolutions, which is helpful for evaluating [parameterizations](https://github.com/pyqg/pyqg/pull/293) :)

pyqg/layered_model.py
```diff
@@ -282,49 +282,49 @@ def _initialize_model_diagnostics(self):
...
+                units='meters second ^-2',
                 dims=('l','k')
         )
 
         self.add_diagnostic('KEspec_modal',
                 description='modal kinetic energy spectra',
                 function= (lambda self:
                     self.wv2*(np.abs(self.phn)**2)/self.M**2 ),
-                units='',
+                units='meters squared second ^-2',
```
rabernat, line 294: To be honest, I have always felt that our usage of units was a little to verbose. I would prefer to just see.
```suggestion
                units='m^2 s^-2',
```

(which is also pint and udunits compatible)

Would you mind doing a global search and replace to fix this everywhere?

pyqg/layered_model.py
```diff
@@ -137,15 +137,15 @@ def __init__(
     def _initialize_stretching_matrix(self):
         """ Set up the stretching matrix """
 
-        self.S = np.zeros((self.nz, self.nz))
+        self.S = np.zeros((self.nz, self.nz)) # 1/m2
```
francispoulin, line 140: It took me a little bit to realize that `1/m2` is 1/metres squared. Would it be worth having `# 1/meter^2` just to be explcit?

Just a thoguht to make it more readable.  

I think this is a great addition by the way but won't do a review as I am slowly getting back into the library.

rabernat (commented):
This looks great!

I have one nit-picking suggestion about unit naming, which you can choose to ignore if you wish.

Thanks so much for your work @asross! 💪

pyqg/layered_model.py
asross: Sure, that sounds good to me! I was just mimicking what was already there.

pyqg/layered_model.py
asross: Yes, good point -- I actually wasn't sure if I was going to commit these comments initially since they were just for my own sanity as I was checking the units of each diagnostic term, but they probably _are_ useful, and it makes sense to give them the right format.

As a side-note, I wish there was some kind of IDE or explicit way to keep track of units statically when programming (in addition to runtime support from libraries like pint).

pyqg/layered_model.py
asross: Updated!

asross:
Thanks again for the helpful suggestions -- I'll plan to merge this soon unless there are more!

PR #286, opened 2022-03-28
allow LayeredModel to run with default arguments

asross:
(Based on <https://pyqg.readthedocs.io/en/latest/examples/layered.html> -- adding these defaults makes it much easier to write tests that invoke LayeredModel without adding a lot of boilerplate.)

### DIFF
--- a/pyqg/layered_model.py
+++ b/pyqg/layered_model.py
@@ -72,8 +72,9 @@ class LayeredModel(model.Model):
     def __init__(
         self,
         beta = 1.5e-11,             # gradient of coriolis parameter
-        nz = 4,                     # number of layers
+        nz = 3,                     # number of layers
         rd = 15000.0,               # deformation radius
+        f = 0.0001236812857687059,  # coriolis parameter [s^-1]
         H = None,                   # layer thickness
         U = None,                   # zonal base state flow
         V = None,                   # meridional base state flow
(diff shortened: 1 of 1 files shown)


pyqg/layered_model.py
```diff
@@ -113,15 +114,24 @@ def __init__(
         self.beta = beta
         self.rd = rd
         self.delta = delta
+
+        if U is None: U = (np.arange(nz) * 0.025)[::-1]
+        if V is None: V = np.zeros(nz)
+        if H is None: H = [500] + [1750 for _ in range(nz-1)]
+        if rho is None: rho = np.arange(nz) * 0.3 + 1025
```
rabernat, line 121: Why not just use default arguments in the constructor?

rabernat (commented):
This is a good idea Andrew. Just one observation...

pyqg/layered_model.py
asross: Hmm -- that would be better in some respects, but using `None` and overwriting in the body of `__init__` allows `LayeredModel` to work no matter what `nz` is passed.

rabernat: gotcha that makes total sense.

asross: Awesome, will plan to merge then.

PR #288, opened 2022-03-29
Add calculation of dissipation spectrum

dante831:
In `qg_model.py`, added calculation of spectral contribution from dissipation, which is represented by the spectral filter in pyqg. The calculation has been tested and it shows an approximate energy closure at the equilibrium stage when adding to the spectral contribution from other terms.


asross:
Looks good to me -- and nice to see the [updated example](https://pyqg--288.org.readthedocs.build/en/288/examples/two-layer.html#Plot-Diagnostics)! I'll plan to merge soon unless anyone else objects.

rabernat:
Thanks for this!

It would be great to have a test for this, beyond just checking that the field is there. Admittedly we have not been very good about this with our existing diagnostics, but it's never too late to start!

That is an optional suggestion, not a requirement. If you just want to move forward at is, that's fine with me.

pyqg/model.py
```diff
@@ -664,6 +664,36 @@ def parameterization_spectrum(m):
...
+                dt3 = 5./12.*m.dt
+            for k in range(m.nz):
+                spectrum[k] += (m.filtr - ones) * (
+                    m.qh[k] + dt1*m.dqhdt[k] + dt2*m.dqhdt_p[k] + dt3*m.dqhdt_pp[k])
+            return -np.real(np.tensordot(m.Hi, np.conj(m.ph) * spectrum, axes = (0, 0))) / m.H / m.dt
+
+        self.add_diagnostic('Dissspec',
+            description='Spectral contribution of filter dissipation',
+            function=dissipation_spectrum,
+            units='',
```
rabernat, line 693: Why are the units blank?

dante831: Added units for Dissspec ("meters squared second ^-3", consistent with the `diagnostics-units` branch)

dante831:
> Thanks for this!
> 
> It would be great to have a test for this, beyond just checking that the field is there. Admittedly we have not been very good about this with our existing diagnostics, but it's never too late to start!
> 
> That is an optional suggestion, not a requirement. If you just want to move forward at is, that's fine with me.

Are tests located in `pyqg/tests/` folder? I guess it would make sense to write a test for Dissspec in file `pyqg/tests/test_diagnostics.py`.

asross:
> Are tests located in `pyqg/tests/` folder? I guess it would make sense to write a test for Dissspec in file `pyqg/tests/test_diagnostics.py`.

Maybe a test that it's only nonzero above the cutoff wavenumber would make sense, though it's true that there aren't unit tests (or units 😃) for other parts of the diagnostics code. But I guess the best time to start is always now.

dante831:
Wrote a test code under `pyqg/tests/test_diagnostics.py` to show that the dissipation contribution is just the difference between the filtered rhs and the unfiltered rhs!

asross:
Looks great to me -- awesome test! I'll plan to merge later today unless @rabernat has any other suggestions.

PR #290, opened 2022-04-05
Automatic process for publishing on release

asross:
This PR should further simplify our publish process, automating it when we create a new tagged GitHub release. It will also resolve the last TODO of https://github.com/pyqg/pyqg/issues/255.


asross:
Interestingly, the most recent publish to TestPyPI failed, but that's because PyPI itself is down: https://twitter.com/PythonStatus/status/1511364423796985860

rabernat (approved):
LGTM!

Issue #292, opened 2022-04-15
Adding a library of parameterizations / new documentation section


asross:
I'm thinking of adding a library of ready-to-use parameterizations, such that you could write, e.g.:

```python
import pyqg

param = pyqg.SmagorinskyParameterization(constant=0.05)
model = pyqg.QGModel(uv_parameterization=param)
```

In addition, I think it might make sense to create a new documentation section where each parameterization is briefly described, we provide instructions for contributing new ones, and include additional documentation, e.g. for https://github.com/pyqg/pyqg/pull/291.

It might also be nice to include an example notebook where we compare all the parameterizations in the library (a la [this nice documentation page from scikit-learn](https://scikit-learn.org/stable/modules/clustering.html#overview-of-clustering-methods)), and potentially include some parameterization evaluation helpers.

Let me know if anyone has any thoughts / thinks this would be useful.

PR #293, opened 2022-04-18
Add a library of example parameterizations

asross:
See https://github.com/pyqg/pyqg/issues/292. ~~Some probably-necessary TODOs before merging:~~
- [x] create a new documentation section (including a guide for contributing parameterizations)
- [x] create an example notebook comparing each parameterization
- [x] integrate with additional work from @dante831

Now that these are done, it would be great to get both review of the code and the docs. In case it's easier to review docs in their fully rendered form, here are the updated sections:
- <https://pyqg--293.org.readthedocs.build/en/293/equations/parameterizations.html>
- <https://pyqg--293.org.readthedocs.build/en/293/api.html#parameterizations>
- <https://pyqg--293.org.readthedocs.build/en/293/examples/comparing_parameterizations.html>


pyqg/qg_diagnostics.py
```diff
@@ -29,7 +30,8 @@ def _initialize_model_diagnostics(self):
                         self.a, self._calc_parameterization_contribution()))
                 ),
             units='m^2 s^-3',
-            dims=('l','k')
+            dims=('l','k'),
+            skip_comparison=True,
```
asross, line 34: I'm not happy about adding more attributes to the diagnostics dictionaries, but I think we can address that as we refactor diagnostics to first-class objects.

pyqg/model.py
```diff
@@ -371,6 +388,129 @@ def stability_analysis(self,bottom_friction=False):
...
+            x1 = getattr(m1, quantity)
+            x2 = getattr(m2, quantity)
+            # Compute the difference in advected quantities
+            forcing = coarsening_function(m1._advect(x1)) - m2.ifft(m2._advect(x2))
+            forcings.append(forcing)
+
+        if return_one:
+            forcings = forcings[0]
+
+        return forcings
```
asross, line 512: This function could go in a different pull request if that makes reviewing easier.

docs/equations/parameterizations.rst
```diff
@@ -0,0 +1,197 @@
...
+:code:`uv_parameterization` arguments:
+
+.. code-block:: python
+
+    param1 = pyqg.Smagorinsky() # targets uv
+    param2 = pyqg.ZannaBolton2020() # also targets uv
+    good_model = pyqg.QGModel(parameterization=param1 + 0.25*param2) # this works!
+
+    param3 = pyqg.BackscatterBiharmonic() # targets q
+    bad_model = pyqg.QGModel(parameterization=param1 + param3) # this will error!
```
rabernat, line 39: I would prefer not to have multiple ways to pass the parameterization. Can we eliminate `parameterization=` and just force the use of `uv_parameterization` or `q_parameterization`?

pyqg/model.py
```diff
@@ -118,6 +121,7 @@ def __init__(
         g= 9.81,                    # acceleration due to gravity
         q_parameterization=None,    # subgrid parameterization in terms of q
         uv_parameterization=None,   # subgrid parameterization in terms of u,v
+        parameterization=None,      # subgrid parameterization (type will be inferred)
```
rabernat, line 124: I'd prefer to drop this option

pyqg/model.py
```diff
@@ -168,25 +172,44 @@ def __init__(
...
+            the model as input and returns a tuple of two :code:`numpy` arrays,
+            each of shape  :code:`(nz, ny, nx)`, to be added to the zonal and
+            meridional velocity derivatives (respectively) at each timestep (by
+            adding their curl to :math:`\partial_t q`).  This can also be used
+            to implemented subgrid forcing parameterizations, but expressed in
+            terms of velocity rather than potential vorticity.
+        parameterization : pyqg.Parameterization
+            An explicit :code:`Parameterization` object representing either a
+            velocity or potential vorticity parameterization, whose type will
+            be inferred.
```
rabernat, line 192: drop

pyqg/model.py
```diff
@@ -168,25 +172,44 @@ def __init__(
...
+        if parameterization is not None:
+            ptype = parameterization.parameterization_type 
+            if ptype == 'uv_parameterization':
+                assert uv_parameterization is None
+                uv_parameterization = parameterization
+            elif ptype == 'q_parameterization':
+                assert q_parameterization is None
+                q_parameterization = parameterization
+            else:
+                raise ValueError(f"unknown parameterization type {ptype}")
```
rabernat, line 211: drop

pyqg/model.py
```diff
@@ -795,3 +948,17 @@ def to_dataset(self):
...
+        PV or velocity, warning if there are both).
+
+        Returns
+        -------
+        parameterization : pyqg.Parameterization or function
+        """
+        if self.q_parameterization and self.uv_parameterization:
+            warnings.warn("Model has multiple parameterizations, "\
+                          "but only returning PV")
+        return self.q_parameterization or self.uv_parameterization
```
rabernat, line 964: drop

pyqg/tests/test_subgrid_forcing.py
```diff
@@ -0,0 +1,68 @@
...
+    Su, Sv, Sq = m.subgrid_forcing(16, ['u','v','q'])
+    assert Su.shape == (2,16,16)
+    assert Sv.shape == (2,16,16)
+    assert Sq.shape == (2,16,16)
+
+    # test for invalid arguments
+    with pytest.raises(AssertionError): m.subgrid_forcing(64)
+    with pytest.raises(AssertionError): m.subgrid_forcing(63)
+    with pytest.raises(AssertionError): m.subgrid_forcing(8, 'qh')
+    with pytest.raises(AssertionError): m.subgrid_forcing(8, ':)')
```
rabernat, line 68: Great test! 🚀

rabernat (commented):
Overall this looks absolutely fantastic. 👏 

My only real comment is that I don't think we should both trying to support the generic `parameterization` argument. We should just have q or uv parameterizations. This will force the user to be more explicit allow us to remove code dealing with exception handling.

pyqg/model.py
```diff
@@ -168,25 +172,44 @@ def __init__(
...
 
         if ny is None:
             ny = nx
         if W is None:
             W = L
 
+        # if an explicit parameterization object was passed without a given
+        # type, infer it from its attributes
+        if parameterization is not None:
+            ptype = parameterization.parameterization_type
```
rabernat, line 203: I think this line will error if you just pass a regular function. better to use
```suggestion
            ptype = getattr(parameterization, "parameterization_type", None)
```

pyqg/parameterizations.py
```diff
@@ -0,0 +1,292 @@
...
+        return f"{self.weight} * {self.param}"
+
+class UVParameterization(Parameterization):
+    """A generic class representing a subgrid parameterization in terms of
+    velocity. Inherit from this to define a new velocity parameterization."""
+
+    @property
+    def parameterization_type(self):
+        ""
+        return 'uv_parameterization'
```
rabernat, line 126:
```suggestion
    parameterization_type(self) = 'uv_parameterization'
```

class attribute is probably simpler here, no?

docs/equations/parameterizations.rst
rabernat: We talked on slack and decided to punt on this suggestion.

pyqg/parameterizations.py
asross: Hmm -- yes, just need to ensure that there are no issues with the `@abstract` `@property` not being implemented

asross: Nevermind, it's fine, I misunderstood ABC

pyqg/model.py
asross: Per an offline discussion, we're keeping this for now, but considering refactoring this to just take a single `parameterizations` iterable.

Issue #299, opened 2022-05-04
Refactoring repetitive code in layered_model.py and qg_model.py


dante831:
Per [comment](https://github.com/pyqg/pyqg/pull/297#issuecomment-1113252846) by @asross, we now have many lines of diagnostic code in `layered_model.py` and `qg_model.py` that are the same. These two classes are also very similar by nature, and so many calculations are the same as well. It might make sense to refactor repetitive code and create a subclass which can be inherited by `QGModel` and `LayeredModel`.

PR #300, opened 2022-05-04
Refactor QG diagnostics to shared class

asross:
Addresses https://github.com/pyqg/pyqg/issues/299. Also renames `paramspec_{apeflux,keflux}` to `paramspec_{APEflux,KEflux}` for consistency.


dante831:
Awesome! 
We could further combine `KEflux_div`, `APEflux_div` in `layered_model.py` and `KEflux`, `APEflux` in `qg_model.py`. The former is mathematically the same, and the latter diagnostic is theoretically the same with difference due to numerical errors. We would want to use the forms in `layered_model.py` as they are more general. 
 If you'd like, we could also assimilate variables in `_calc_derived_fields` function because they are used in `qg_diagnostics.py`. The common variables after the above merge will be `p`, `xi`, `Jpxi`, `Jq`,  `Sph`, `Sp`, `JSp`. `Jptpc` in `qg_model.py` could be removed.

asross:
That sounds great -- maybe we can merge this first, then make those additional refactors?

asross:
@dante831 going to merge this now to enable some other changes, but looking forward to making/reviewing those other refactors in a future PR :)

Issue #275, opened 2022-01-21
Continued issues with `calc_ispec`


asross:
Per discussions with @Pperezhogin, we think there might still be errors in the [`calc_ispec`](https://github.com/pyqg/pyqg/blob/723d94ad1573dcf28a134dd6fb49fa966f8ef41a/pyqg/diagnostic_tools.py#L52-L85) helper (even after [my PR](https://github.com/pyqg/pyqg/pull/273) which attempted to address [a more glaring issue](https://github.com/pyqg/pyqg/issues/272)). However, we're not totally sure how to fix it in a way that fits the other requirements of the library, and especially how to create a good unit test to verify it. This is a pretty important issue for us to fix, however, since other analyses depend on it.

Specific issues we've been discussing include:
- the minimum radial wavenumber probably should be `min(dk, dl)` rather than 1.5x (its current value) or 0.5x (its previous value)
- the normalization factor defined [here](https://github.com/pyqg/pyqg/pull/273) might be incorrect, and it seems like we should be instead using the Parseval identity (correct me if I'm mischaracterizing this @Pperezhogin)
- there might be double-counting of quantities at particular wavenumbers, potentially because [the intervals we sum over](https://github.com/pyqg/pyqg/blob/723d94ad1573dcf28a134dd6fb49fa966f8ef41a/pyqg/diagnostic_tools.py#L81) might not be closed (due to the `<=` rather than `<`) and potentially because we might not be handling the `k` and `l` difference (i.e. some with complex conjugates, some unique) correctly, though this is a bit outside of my area of familiarity

So, just wanted to bring this issue up and see if anyone has thoughts on how to fix it -- ideally with a unit test with some complex array where we have ground-truth knowledge about the complete isotropic spectrum and can verify that this method returns it.

rabernat:
Thanks for raising this issue Andrew! The isotropic spectral diagnostics were originally implemented by @cesar-rocha, who is no longer really involved in this project.

In our spectral analysis package, we have an "isotropize" routine which may be a useful reference here

https://github.com/xgcm/xrft/blob/023c86e4e3c8d331604ab7ec1027f04bcf7ae0eb/xrft/xrft.py#L996-L1052

rabernat:
Also related:
- https://github.com/xgcm/xrft/issues/145

MFJansen:
A couple of quick comments on this:
[...]

PR #301, opened 2022-05-06
Update calc_ispec to resolve issues and (sort of) preserve units

asross:
Hopefully resolves https://github.com/pyqg/pyqg/issues/275 (using @Pperezhogin's implementation times a normalization factor to preserve units), but it would be great to get some detailed feedback on this.

Effect of these changes:
- Fixes the minimum radial wavenumber
- Fixes a potentially problematic (though probably fine) double-counting issue
- Allows for more options re: averaging
- Changes the normalization of the isotropic spectrum, so the values aren't many orders of magnitude different than the those of the zonal or meridional spectrum (but I'm not 100% sure if multiplying by `dkr` or `kmin` is more correct)

Illustration:
image

### DIFF
--- a/pyqg/tests/test_diagnostic_tools.py
+++ b/pyqg/tests/test_diagnostic_tools.py
@@ -46,17 +46,14 @@ def test_calc_ispec_units():
         if m1.diagnostics[diagnostic]['dims'] == ('l','k'):
             spec1 = m1.diagnostics[diagnostic]['function'](m1)
             spec2 = m2.diagnostics[diagnostic]['function'](m2)
-            spec1x = spec1.sum(axis=0)
-            spec1y = spec1.sum(axis=1)
-            spec2x = spec2.sum(axis=0)
-            spec2y = spec2.sum(axis=1)
-            _, spec1r = calc_ispec(m1, spec1)
-            _, spec2r = calc_ispec(m2, spec2)
+            k1r, spec1r = calc_ispec(m1, spec1)
+            k2r, spec2r = calc_ispec(m2, spec2)
 
             scales = [
-                np.abs(spec).mean()
-                for spec in [spec1x, spec1y, spec1r,
-                             spec2x, spec2y, spec2r]
+                np.abs(spec1).sum(),
+                np.abs(spec2).sum(),
+                np.trapz(np.abs(spec1r), k1r),
+                np.trapz(np.abs(spec2r), k2r)
             ]
 
             if max(scales) == 0:
@@ -64,8 +61,8 @@ def test_calc_ispec_units():
 
             scale_ratio = max(scales) / min(scales)
 
-            assert scale_ratio < 4, \
-                f"calc_ispec should preserve units of {diagnostic}"
-            assert scale_ratio > 0.25, \
-                f"calc_ispec should preserve units of {diagnostic}"
+            assert scale_ratio < 10, \
+                f"calc_ispec should have correct units for {diagnostic}"
+            assert scale_ratio > 0.1, \
+                f"calc_ispec should have correct units for {diagnostic}"
 
--- a/pyqg/tests/test_diagnostic_tools.py
+++ b/pyqg/tests/test_diagnostic_tools.py
@@ -1,9 +1,11 @@
 import numpy as np
 import pyqg
 import pytest
+import os
+import pickle
 from pyqg.diagnostic_tools import calc_ispec
 
-def test_calc_ispec():
+def test_calc_ispec_peak():
     # Create a radial sine wave spiraling out from the center of the model's
     # spatial field (with a given frequency)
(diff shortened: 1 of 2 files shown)


dante831:
Thanks for fixing this! I have a few suggested modifications after having done some tests on this. 

- Manually multiply by 2 onto `phr`, so that the normalization is signal2d.var() = phr.sum() * (kr[1] - kr[0]). This is due to the fact that the model uses only half of the coefficients, but really we want the spectrum of the whole field, not half of its strength. 
- Fix the definition of `fkr`, so that the bin edges are not counted twice. 
- Put `phr = phr * kmin` inside `if nd_wavenumber:` clause. This ensures that Parseval's theorem is satisfied, i.e., the integral of `phr` along `kr` equals to the total variance. One can check by comparing `var_dens.sum()*2` and `np.trapz(phr, kr)`. 

In lieu of these modifications, the zonal or meridional spectrum that you showed in the plot above should probably divide by `kmin` to ensure that it's also "density", which means that its integral is the total energy in the domain. Then, the units of KEspec should be "m^3 s^-2" since it's energy divided by the units of wavenumber. 

The code after modification is below: 

```python
def calc_ispec(model, _var_dens, averaging = True, truncate=True, nd_wavenumber=False, nfactor = 1):
    """Compute isotropic spectrum `phr` from 2D spectrum of variable signal2d.

    Parameters
    ----------
    model : pyqg.Model instance
        The model object from which `var_dens` originates
    
    var_dens : squared modulus of fourier coefficients like this:
        np.abs(signal2d_fft)**2/m.M**2

    averaging: If True, spectral density is estimated with averaging over circles,
        otherwise summation is used and Parseval identity holds

    truncate: If True, maximum wavenumber corresponds to inner circle in Fourier space,
        otherwise - outer circle
    
    nd_wavenumber: If True, wavenumber is nondimensional: 
        minimum wavenumber is 1 and corresponds to domain length/width,
        otherwise - wavenumber is dimensional [m^-1]

    nfactor: width of the bin in sqrt(dk^2+dl^2) units

    Returns
    -------
    kr : array
        isotropic wavenumber
    phr : array
        isotropic spectrum

    Normalization:
    signal2d.var() = phr.sum() * (kr[1] - kr[0])
    """

    # account for complex conjugate
    var_dens = np.copy(_var_dens)
    var_dens[...,0] /= 2
    var_dens[...,-1] /= 2

    ll_max = np.abs(model.ll).max()
    kk_max = np.abs(model.kk).max()

    if truncate:
        kmax = np.minimum(ll_max, kk_max)
    else:
        kmax = np.sqrt(ll_max**2 + kk_max**2)
    
    kmin = np.minimum(model.dk, model.dl)

    dkr = np.sqrt(model.dk**2 + model.dl**2) * nfactor

    # left border of bins
    kr = np.arange(kmin, kmax, dkr)
    
    phr = np.zeros(kr.size)

    for i in range(kr.size):
        if i == kr.size-1:
            fkr = (model.wv>=kr[i]) & (model.wv<=kr[i]+dkr)
        else:
            fkr = (model.wv>=kr[i]) & (model.wv<kr[i]+dkr)
        if averaging:
            phr[i] = var_dens[fkr].mean() * (kr[i]+dkr/2) * pi / (model.dk * model.dl)
        else:
            phr[i] = var_dens[fkr].sum() / dkr

        phr[i] *= 2 # include full circle
    
    # convert left border of the bin to center
    kr = kr + dkr/2

    # convert to non-dimensional wavenumber 
    # preserving integral over spectrum
    if nd_wavenumber:
        kr = kr / kmin
        phr = phr * kmin

    return kr, phr
```

asross:
Awesome, thanks for this! I'll apply your suggestion, wait a little bit for any more comments, and then plan to merge :)

pyqg/tests/test_diagnostic_tools.py
```diff
@@ -21,3 +23,46 @@ def test_calc_ispec():
...
+
+            if max(scales) == 0:
+                continue
+
+            scale_ratio = max(scales) / min(scales)
+
+            assert scale_ratio < 10, \
+                f"calc_ispec should have correct units for {diagnostic}"
+            assert scale_ratio > 0.1, \
+                f"calc_ispec should have correct units for {diagnostic}"
```
asross, line 67: Let me know if this is the right assertion @dante831 -- I think I'm testing that the integral of the l,k spectra match the integral of the radial spectra, at least to one order of magnitude.

dante831:
Okay I've updated calc_ispec() such that its integral is the same as `signal2d.var()` instead of `signal2d.var()/2`, and updated code such that rtol can be `1e-5` instead of `1e-2`. I also added another testing routine `test_calc_ispec_sum()` to explicitly compare half plane and full plane, and check the applicability of Parseval's theorem on the radial spectrum, half plane and the full plane.

Issue #240, opened 2021-05-06
Provide base class for kernels


rabernat:
As discussed in today's meting, this will make it easier to implement new kernels (e.g. #167, #215).

asross:
I've been thinking about this as well. Were you thinking something along these lines?
- `pyqg.Model` stops being a _subclass_ of `pyqg.Kernel`, but rather accepts a `kernel` argument, which defaults to something like `pyqg.CythonFFTWKernel`
- We create a base `pyqg.Kernel` class that the others inherit from
- For backwards compatibility, we create `@property` getters and setters in `pyqg.Model` which allow all main kernel variables to be read and updated from the model level (rather than requiring users to do, e.g., `model.kernel.q` or `model.kernel.rek`)

rabernat:
I agree with all three of your design suggestions. 👍 

I think the hard part will be defining the `Kernel` interface. But it is doable.

asross:
I think I was able to extract the kernel fairly cleanly: https://github.com/pyqg/pyqg/pull/303

Issue #241, opened 2021-05-06
Implement jax kernel


rabernat:
Requires #240.

cmdupuis3:
While working on some sketches for imlementing the jax kernel, I learned that most of jax is designed to be purely functional. This poses a problem for implementation because the original kernel is designed as a state machine. For example, if we have

```
class PseudoSpectralKernel:

    def __init__(self, nz, ny, nx):
        self.q = jnp.zeros((self.nz, self.ny, self.nx), jnp.float32)
```

a stateless implementation of `self.q` would imply that `self.q` is constant. This doesn't mean we _can't_ overwrite `self.q`, but it does mean that we cannot do any reassignment without overwriting the entire array. This should be avoided unless we can use array operations that cover all the dimensions, so whatever is assigned to `self.q` needs to be written as an applicative or something broadcastable.

Also note another consequence: initializing with `zeros` won't make sense in most cases because it should be completely overwritten.

A redesign of the kernel should probably reflect the principles jax is based on, and to that end, a good design could be the _static objects + functions_ design common in functional languages with OO capabilities. In other words, we can have a struct that stores an individual model state, and passing it to some function(s) will generate the next model state.

Tagging @asross re:slack

cmdupuis3:
Advised to delete `_do_external_forcing`, which will streamline calculation of dq/dt

cmdupuis3:
Note: default Jax FFTs run in single precision, while the current tests apparently expect double precision.

PR #303, opened 2022-05-17
WIP: extracting kernel to a separate object using delegation

asross:
This pull requests makes `pyqg.Model` a top-level class, invoking `pyqg.PseudoSpectralKernel` by dependency injection rather than inheritance.

This change will allow us to pass in alternative kernels (e.g. jax or numpy without Cython), addressing https://github.com/pyqg/pyqg/issues/240.

Note that as currently implemented, I'm using the [delegate-property](https://github.com/dscottboggs/python-delegate) package, which adds a dependency. Because it's is a [fairly short library](https://github.com/dscottboggs/python-delegate/blob/master/delegate/__init__.py), I could copy it over if we think it might cause problems, or at least lock the version in `requirements.txt`.

### DIFF
--- a/pyqg/layered_model.py
+++ b/pyqg/layered_model.py
@@ -118,8 +118,8 @@ class LayeredModel(qg_diagnostics.QGDiagnostics):
         if H is None: H = [500] + [1750 for _ in range(nz-1)]
         if rho is None: rho = np.arange(nz) * 0.3 + 1025
 
-        self.Ubg = np.array(U)
-        self.Vbg = np.array(V)
+        self._U = U
+        self._V = V
         self.Hi = np.array(H)
         self.rhoi = np.array(rho)
 
@@ -178,6 +178,9 @@ class LayeredModel(qg_diagnostics.QGDiagnostics):
 
     def _initialize_background(self):
         """Set up background state (zonal flow and PV gradients)."""
+
+        self.Ubg = np.array(self._U)
+        self.Vbg = np.array(self._V)
 
         self.H = self.Hi.sum()
 
--- a/pyqg/model.py
+++ b/pyqg/model.py
@@ -3,6 +3,7 @@ from numpy import pi
 import logging
 import warnings
 import inspect
+from delegate import delegate
 
 from .errors import DiagnosticNotFilledError
 from .kernel import PseudoSpectralKernel, tendency_forward_euler, tendency_ab2, tendency_ab3
@@ -19,7 +20,26 @@ try:
 except ImportError:
     pass
 
-class Model(PseudoSpectralKernel):
+kernel_attrs = [
+    'nx', 'ny', 'nz', 'nl', 'nk', 'a', 'kk', 'll', 
+    'q', 'qh', 'ph', 'u', 'v',
+    'Ubg', 'Qy', 'ufull', 'vfull',
+    'uh', 'vh', 'rek', 't', 'tc', 'dt',
+    'uv_parameterization', 'q_parameterization',
+     'fft', 'ifft', 'filtr', 'ablevel',
+    'dqdt', 'dqhdt', 'dqhdt_p', 'dqhdt_pp', 'dqh', 'duh', 'dvh',
+    'uq', 'vq',
+    '_invert',
+    '_forward_timestep',
+    '_do_advection',
+    '_do_friction',
+    '_do_q_subgrid_parameterization',
+    '_do_uv_subgrid_parameterization',
+    '_ik', '_il',
+]
+
+@delegate(*kernel_attrs, to='kernel')
+class Model:
     """A generic pseudo-spectral inversion model.
 
     Attributes
@@ -100,6 +120,7 @@ class Model(PseudoSpectralKernel):
 
     def __init__(
         self,
+        kernel=PseudoSpectralKernel,
         # grid size parameters
         nz=1,
         nx=64,                     # grid resolution
@@ -213,9 +234,11 @@ class Model(PseudoSpectralKernel):
 
         # TODO: be more clear about what attributes are cython and what
         # attributes are python
-        PseudoSpectralKernel.__init__(self, nz, ny, nx, ntd,
-                has_q_param=int(q_parameterization is not None),
-                has_uv_param=int(uv_parameterization is not None))
+        self.kernel = kernel(nz, ny, nx,
+                q_parameterization,
+                uv_parameterization,
+                ntd,
+        )
 
         self.L = L
         self.W = W
@@ -240,10 +263,6 @@ class Model(PseudoSpectralKernel):
         if f:
             self.f = f
             self.f2 = f**2
-
-        # optional subgrid parameterizations
-        self.q_parameterization = q_parameterization
-        self.uv_parameterization = uv_parameterization
 
         # TODO: make this less complicated!
         # Really we just need to initialize the grid here. It's not necessary
--- a/pyqg/qg_model.py
+++ b/pyqg/qg_model.py
@@ -86,19 +86,19 @@ class QGModel(qg_diagnostics.QGDiagnostics):
         U2 : number
             Lower layer flow. Units: meters seconds :sup:`-1`
         """
+        if 'nz' in kwargs:
+            del kwargs['nz']
+
+        self.delta = delta
+        self.Hi = np.array([ H1, H1/delta])
 
         # physical
         self.beta = beta
         #self.rek = rek
         self.rd = rd
-        self.delta = delta
-        self.Hi = np.array([ H1, H1/delta])
         self.U1 = U1
         self.U2 = U2
         #self.filterfac = filterfac
-
-        if 'nz' in kwargs:
-            del kwargs['nz']
 
         super().__init__(nz=2, **kwargs)
 
--- a/pyqg/bt_model.py
+++ b/pyqg/bt_model.py
@@ -43,15 +43,13 @@ class BTModel(model.Model):
         self.Hi = np.array(H)[np.newaxis,...]
         self.U = U
 
-        self.nz = 1
-
         # deformation wavenumber
         if rd:
             self.kd2 = rd**-2
         else:
             self.kd2 = 0.
 
-        super().__init__(**kwargs)
+        super().__init__(nz=1, **kwargs)
 
         # initial conditions: (PV anomalies)
         self.set_q(1e-3*np.random.rand(1,self.ny,self.nx))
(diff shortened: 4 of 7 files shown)


pyqg/model.py
```diff
@@ -19,7 +21,26 @@
...
+    '_invert',
+    '_forward_timestep',
+    '_do_advection',
+    '_do_friction',
+    '_do_q_subgrid_parameterization',
+    '_do_uv_subgrid_parameterization',
+    '_ik', '_il',
+]
+
+@delegate(*kernel_attrs, to='kernel')
```
asross, line 42: So I think the delegation here is good, but any thoughts on how to organize the list of attributes we're delegating (or if there are any we should remove the need for)?

pyqg/model.py
```diff
@@ -213,9 +235,11 @@ def __init__(
 
         # TODO: be more clear about what attributes are cython and what
         # attributes are python
-        PseudoSpectralKernel.__init__(self, nz, ny, nx, ntd,
-                has_q_param=int(q_parameterization is not None),
-                has_uv_param=int(uv_parameterization is not None))
+        self.kernel = kernel(nz, ny, nx,
+                q_parameterization,
+                uv_parameterization,
+                ntd,
```
asross, line 241: One issue -- `ntd` (number of threads) is not going to be an attribute that every kernel needs (or accepts). We could get around this by passing an optional `kernel_kwargs` argument but that seems questionable.

Normally, with dependency injection, we pass in a fully initialized object, but here we're passing in a class. That also seems questionable. Hmm.

rabernat:
Two options if we want to use delegate-property:
- "Vendor" it (just copy the full code here)
- Create a conda forge recipe for it

pyqg/delegate.py
```diff
@@ -0,0 +1,49 @@
+# Taken from https://github.com/dscottboggs/python-delegate
```
rabernat, line 1: We need to repackage their license and copyright information.

pyqg/model.py
rabernat: How do you imagine the api will work once we have multiple kernels. How will the user specify which kernel they want to use at runtime?

I would prefer to be able to write

```
model = QGModel(kernel="jax", kernel_kwargs={'foo': 'special jax argument'})
```

asross: Yeah, I was thinking something like that -- it would support both
```python
model = pyqg.QGModel(kernel="jax", kernel_kwargs={"foo": "special jax argument"})
```
or
```python
model = pyqg.QGModel(kernel=pyqg.JaxKernel, kernel_kwargs={"foo": "special jax argument"})
```

pyqg/model.py
```diff
@@ -19,7 +21,26 @@
...
+     'fft', 'ifft', 'filtr', 'ablevel',
+    'dqdt', 'dqhdt', 'dqhdt_p', 'dqhdt_pp', 'dqh', 'duh', 'dvh',
+    'uq', 'vq',
+    '_invert',
+    '_forward_timestep',
+    '_do_advection',
+    '_do_friction',
+    '_do_q_subgrid_parameterization',
+    '_do_uv_subgrid_parameterization',
+    '_ik', '_il',
```
asross, line 39: TODO: make these attributes non-private or remove them

pyqg/model.py
asross: There are fewer attributes being delegated now, so maybe this is ok :)

asross:
Sorry for all the draft-no-draft conversions -- keep forgetting to re-compile before running tests :) Current issue is just figuring out how the Cython kernel inherits from a Python class.

asross:
> Sorry for all the draft-no-draft conversions -- keep forgetting to re-compile before running tests :) Current issue is just figuring out how the Cython kernel inherits from a Python class.

It looks like this is actually impossible! We might still eventually add inheritance, but just for Python kernels (e.g. numpy and jax).

pyqg/model.py
```diff
@@ -100,6 +107,8 @@ class Model(PseudoSpectralKernel):
 
     def __init__(
         self,
+        kernel=kernels.CythonFFTWKernel,
+        kernel_kwargs={},
```
rabernat, line 111: These need to be documented in the docstring

pyqg/model.py
```diff
@@ -550,6 +552,24 @@ def _step_forward(self):
...
+        self.kernel.do_friction()
+
+    def _do_q_subgrid_parameterization(self):
+        self.kernel.do_q_subgrid_parameterization()
+
+    def _do_uv_subgrid_parameterization(self):
+        self.kernel.do_uv_subgrid_parameterization()
+
+    def _forward_timestep(self):
+        self.kernel.forward_timestep()
```
rabernat, line 572: This feels like a code smell. Can we avoid wrapping all these methods? Could we use delegate here?

rabernat, line 572: This feels like a code smell. Can we avoid wrapping all these methods? Could we use delegate here?

pyqg/qg_model.py
```diff
@@ -86,20 +86,20 @@ def __init__(
         U2 : number
             Lower layer flow. Units: meters seconds :sup:`-1`
         """
+        if 'nz' in kwargs:
+            del kwargs['nz']
```
rabernat, line 90: Why? Can we avoid this? Can we just remove the `**kwargs` argument completely?

rabernat, line 90: Why? Can we avoid this? Can we just remove the `**kwargs` argument completely?

pyqg/layered_model.py
```diff
@@ -118,8 +118,8 @@ def __init__(
         if H is None: H = [500] + [1750 for _ in range(nz-1)]
         if rho is None: rho = np.arange(nz) * 0.3 + 1025
 
-        self.Ubg = np.array(U)
-        self.Vbg = np.array(V)
+        self._U = U
+        self._V = V
```
rabernat, line 122: Can you explain why you moved this?

rabernat, line 122: Can you explain why you moved this?

rabernat (commented):
This is looking good. 👍 

It is also turning up some weird and inconsistent patterns within the pyqg code. Can we use this as a chance to clean things up a bit?

A few comments...

pyqg/model.py
asross: I agree that's better. The issue is that, having removed underscores from the corresponding `kernel` method names, then we end up needing to remove underscores from the `pyqg.Model` method names as well, which means they're now part of the public API (and it also breaks some of my code which calls the private methods anyway 😬).

A terse and backwards-compatible alternative I spent a little while starting to implement was updating `delegate` to support a prefix option:

```python
@delegate(*private_kernel_attrs, to='kernel', prefix='_')
@delegate(*public_kernel_attrs, to='kernel')
class Model:
    # ...
```

This is relatively simple in principle, though it requires significant changes to the `delegate` decorator to allow calls to stack / prevent one call from overwriting the other. Because of that, I opted for the smelly solution. However, with your comment, I'm considering updating the decorator again...

pyqg/qg_model.py
asross: This got added in an earlier PR to facilitate reinitializing a new model with the same configuration parameters, but we can avoid it in a much better way. I'll make sure it's removed. I don't think removing `**kwargs` makes as much sense though, since we have them in every other model and it's nice to avoid needing to respecify all the parameters of the parent class.

pyqg/layered_model.py
asross: This needed to get moved up because we can't set `Ubg` before calling `super`, since `Ubg` is now a property of the kernel. However, `super` calls `_initialize_background` which then calls `_initialize_stretching_matrix` which requires `Ubg` to be defined. So, following `QGModel`, I moved the actual setting of `Ubg` to `_initialize_background` (which makes sense). The issue is that we need some intermediate variable to store the array.

pyqg/qg_model.py
asross: Alright, so it looks like the `delegate` changes play poorly here with the `_config` method (which is used in that re-initialization procedure I just alluded to), but also that method just needs a lot more testing. I might add some additional tests to this pull request because the existing ones did not pick up the difference.

asross: Ok, so we can resolve that issue by removing the spurious `__init__` redefinition from delegate, but this is turning into a bit of a can of worms. I think we should actually start off by removing the existing `subgrid_forcing` method and all associated methods (e.g. `_config`), maybe treating it as a patch-level update because it's basically a bug fix. Then we rebase this on top of that change and everything is much simpler. We can revisit subgrid forcing / everything else in a subsequent PR.

asross: Here's a PR to do that: https://github.com/pyqg/pyqg/pull/308

asross: Alright, this is now done.

pyqg/model.py
asross: Alright, I updated it, and also made a pull request to the original library in case they want to allow this functionality (https://github.com/dscottboggs/python-delegate)

pyqg/model.py
asross: I've done the bare minimum to document this now (some description in the `pyqg.Model` docstring), but I'm realizing that for scalability, we probably need to add kernels to the API and give them their own documentation (which users would reference to figure out `kernel_kwargs`).

pyqg/model.py
asross: However, I do think this syntax is a bit weird, since it's more normal to do dependency injection with initialized objects:

```python
model = pyqg.QGModel(kernel=pyqg.CythonFFTWKernel(fftw_num_threads=2)
model = pyqg.QGModel(kernel=pyqg.JaxKernel(foo='bar'))
```

The problem with this is that we'd need to start passing `nx`/`ny`/`nz`/other arguments to the kernel rather than the model, which is could lead to redundancies and definitely isn't backwards compatible.

pyqg/model.py
cmdupuis3: From what I can see, I think this can be resolved by composing most of the kernel functions. I think `Kernel._forward_timestep` is the only one that couldn't be, atm, but only because `_calc_diagnostics` is in the way.

I'm really wary of solving the Kernel/Model inheritance problem with delegation, it takes away a lot of the guarantees I need to be able to implement Jax #241 in a reasonable way. Basically Kernel would become part of Model, which means I can't know if the Kernel arrays are being messed with outside of Kernel. I think if we compose all the kernel functions, that would alleviate this problem on my end, though.

PR #308, opened 2022-05-19
Remove subgrid_forcing method for now

asross:
See https://github.com/pyqg/pyqg/pull/303#discussion_r876393420.


pyqg/model.py
```diff
@@ -394,129 +394,6 @@ def stability_analysis(self,bottom_friction=False):
...
-        Returns
-        -------
-        model : Model
-            Newly initialized model with same or overridden parameters.
-        """
-        config = self._config
-        config.update(kw)
-        return self.__class__(**config)
-
-    def subgrid_forcing(self,
```
rabernat, line 428: Instead of just removing, I would do this
```python
    def subgrid_forcing(self, **kwargs):
        raise ValueError(
            "This method was found to be buggy and removed for now. "
            "It will be re-added in a future release. "
            "See https://github.com/pyqg/pyqg/pull/308 for details. "
        )
```

PR #309, opened 2022-05-19
ADM and Reynolds stress

Pperezhogin:
Added building blocks of deconvolution parameterizations:
- Gaussian filter
- Approximate deconvolution procedure for estimation of subfilter fluctuations (**deconvolve**)
- Approximate deconvolution model estimating subfilter stress
- Reynolds stress model (Perezhogin Glazunov 2021, preprint)

### DIFF
--- a/pyqg/parameterizations.py
+++ b/pyqg/parameterizations.py
@@ -286,3 +286,78 @@ class ZannaBolton2020(UVParameterization):
 
     def __repr__(self):
         return f"ZannaBolton2020(κ={self.constant:.2e})"
+
+def Gaussian_Filter(w, m, FGR):
+    w_fft = m.fft(w)
+    Delta = m.dx * FGR
+    k2 = m.k**2 + m.l**2
+    G = np.exp(-k2*Delta**2/24)
+    return m.ifft(G * w_fft)
+
+def deconvolve(w, m, FGR, order):
+    wd = w
+    r = w
+    for j in range(order):
+        r = r - Gaussian_Filter(r, m, FGR)
+        wd = wd + r
+    return wd
+
+def SFS(q, u, v, m, FGR):
+    qf = Gaussian_Filter(q, m, FGR)
+    uf = Gaussian_Filter(u, m, FGR)
+    vf = Gaussian_Filter(v, m, FGR)
+    quf = Gaussian_Filter(q*u, m, FGR)
+    qvf = Gaussian_Filter(q*v, m, FGR)
+    SFSu = qf*uf - quf
+    SFSv = qf*vf - qvf
+    return SFSu, SFSv 
+
+class ADM(QParameterization):
+    def __init__(self, FGR=2, order=5):
+        self.FGR = FGR
+        self.order = order   
+    
+    def __call__(self, m):
+        qd = deconvolve(m.q, m, self.FGR, self.order)
+        ud = deconvolve(m.u, m, self.FGR, self.order)
+        vd = deconvolve(m.v, m, self.FGR, self.order)
+        
+        SFSu, SFSv = SFS(qd, ud, vd, m, self.FGR)
+        
+        ik = m.k * 1j
+        il = m.l * 1j
+        
+        real = lambda q: q if q.shape == m.q.shape else m.ifft(q)
+        spec = lambda q: q if q.shape != m.q.shape else m.fft(q)
+        ddx = lambda q: real(ik * spec(q))
+        ddy = lambda q: real(il * spec(q))
+        
+        return ddx(SFSu) + ddy(SFSv)
+
+    def __repr__(self):
+        return f'ADM, FGR={self.FGR}, order={self.order}'
+
+class Reynolds_stress(QParameterization):
+    def __init__(self, FGR=2, Csim=12):
+        self.FGR = FGR
+        self.Csim = Csim 
+    
+    def __call__(self, m):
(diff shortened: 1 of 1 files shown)


pyqg/parameterizations.py
```diff
@@ -286,3 +286,78 @@ def __call__(self, m):
...
+        spec = lambda q: q if q.shape != m.q.shape else m.fft(q)
+        ddx = lambda q: real(ik * spec(q))
+        ddy = lambda q: real(il * spec(q))
+        
+        return ddx(SFSu) + ddy(SFSv)
+
+    def __repr__(self):
+        return f'ADM, FGR={self.FGR}, order={self.order}'
+
+class Reynolds_stress(QParameterization):
```
asross, line 340: Also for consistency, could we rename this to `ReynoldsStress`?

rabernat: We should actually set up automatic linting (black and flake8) for the entire project! 🙃  That way we can avoid nit picking style issues on PRs.

asross, line 340: Should add docstrings giving a brief explanation of the parameterization with some latex, following the same pattern as the other parameterizations.

pyqg/parameterizations.py
```diff
@@ -286,3 +286,78 @@ def __call__(self, m):
...
+        
+        ik = m.k * 1j
+        il = m.l * 1j
+        
+        real = lambda q: q if q.shape == m.q.shape else m.ifft(q)
+        spec = lambda q: q if q.shape != m.q.shape else m.fft(q)
+        ddx = lambda q: real(ik * spec(q))
+        ddy = lambda q: real(il * spec(q))
+        
+        return self.Csim*(ddx(SFSu) + ddy(SFSv))
```
asross, line 360: So, the `real`/`spec`/`ddx`/`ddy` methods might be overkill for now. How about we replace everything after line 350 with:
```python
SFSu_x = m.ifft(m.ik * m.fft(SFSu))
SFSv_y = m.ifft(m.il * m.fft(SFSv))
return self.Csim * (SFSu_x + SFSu_y)
```
and similar for the ADM parameterization?

asross: (I guess we could also add `ddx` and `ddy` as instance methods to `pyqg.Model`, and then we could write just one line, `return self.Csim * (m.ddx(SFSu) + m.ddy(SFSv)`)

pyqg/parameterizations.py
```diff
@@ -286,3 +286,78 @@ def __call__(self, m):
...
+
+    def __repr__(self):
+        return f'Reynolds, FGR={self.FGR}, Csim={self.Csim}'
```
asross, line 363: Sorry to be so nitpicky, but the other parameterizations have a slightly different format for their `__repr__` methods, and it would be nice to be consistent (so they look similar on automatically generated plot labels).

Could we change this to:
```suggestion
        return f'ReynoldsStress(FGR={self.FGR}, Csim={self.Csim})'
```

asross, line 363: Sorry to be so nitpicky, but the other parameterizations have a slightly different format for their `__repr__` methods, and it would be nice to be consistent (so they look similar on automatically generated plot labels).

Could we change this to:
```suggestion
        return f'ReynoldsStress(FGR={self.FGR}, Csim={self.Csim})'
```

asross:
Overall, this looks fantastic, just need to add docstrings to the parameterizations / methods so the API docs get automatically updated (see [here](https://pyqg--309.org.readthedocs.build/en/309/api.html#pyqg.parameterizations.ADM) for what they currently look like with this PR).

Additionally, it's important to add a test to [`pyqg/tests/test_parameterizations.py`](https://github.com/pyqg/pyqg/blob/master/pyqg/tests/test_parameterizations.py) to ensure these can be evaluated. This would just look like:
```python
def test_adm(model):
    adm = pyqg.ADM()
    dq = adm(model)

def test_reynolds(model):
    reynolds = pyqg.ReynoldsStress()
    dq = reynolds(model)
```

pyqg/parameterizations.py
```diff
@@ -286,3 +286,78 @@ def __call__(self, m):
...
+
+def SFS(q, u, v, m, FGR):
+    qf = Gaussian_Filter(q, m, FGR)
+    uf = Gaussian_Filter(u, m, FGR)
+    vf = Gaussian_Filter(v, m, FGR)
+    quf = Gaussian_Filter(q*u, m, FGR)
+    qvf = Gaussian_Filter(q*v, m, FGR)
+    SFSu = qf*uf - quf
+    SFSv = qf*vf - qvf
+    return SFSu, SFSv
```
asross, line 313: These methods should probably get [numpy-style docstrings](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html#example-numpy) as well -- e.g.:
```
def SFS(q, u, v, m, FGR):
  """Computes the subfilter stress-or-something-or-other of the quantity :code:`q` 
  based on velocities :code:`u` and :code:`v`, a model :code:`m`, and a filter to
  grid ratio :code:`FGR`.

  Parameters
  ----------
    q : array
       The quantity for which we're computing subfilter stress (?)

    ...

  Returns
  -------
    SFS : tuple of two arrays
      the thingy
  """
```

pyqg/parameterizations.py
asross: Yeah, agreed -- sorry if this review is very nit-picky!

PR #313, opened 2022-09-13
Adding the option of a stochastic ring forcing

roxyboy:
A PR to add the option to force the two-layer QG model with a stochastic ring in wavenumber. 
The configuration is documented [here](https://doi.org/10.31223/X5C063).
The idea was to force the model with stochastic noise associated with spatial scales around the Rossby radius.

### DIFF
--- a/pyqg/tests/test_parameterizations.py
+++ b/pyqg/tests/test_parameterizations.py
@@ -49,5 +49,16 @@ def test_addition_and_scaling(model, rtol=1e-11):
     with pytest.raises(AssertionError):
         back + smag
 
+@pytest.fixture(params=[QG])
+def model(request):
+    klass = request.param()
+    model = klass()
+    model._step_forward()
+    return model
+
+def test_ring(model):
+    ring = pyqg.RingForcing()
+    dq = ring(model)
+
 if __name__ == "__main__":
     unittest.main()
--- a/pyqg/tests/test_parameterizations.py
+++ b/pyqg/tests/test_parameterizations.py
@@ -49,13 +49,6 @@ def test_addition_and_scaling(model, rtol=1e-11):
     with pytest.raises(AssertionError):
         back + smag
 
-@pytest.fixture(params=[QG])
-def model(request):
-    klass = request.param()
-    model = klass()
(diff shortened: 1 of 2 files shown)


asross:
Sorry for the very late reply! This looks great, but do you think you could flesh out the docstring a little bit so that the resulting documentation contains a link to your paper and a brief outline of the parameterization's mathematical formulation? See e.g. the [Smagorinsky parameterization's documentation](https://pyqg.readthedocs.io/en/latest/api.html#pyqg.parameterizations.Smagorinsky), generated from [this docstring](https://github.com/pyqg/pyqg/blob/a8e1eb957d1c3a7fd72b362169e84d5889420268/pyqg/parameterizations.py#L133-L153).

asross:
(Also overall this looks great, just need to fix that error and add a test!)

roxyboy:
@asross Thank you for the feedback. Could you let me know if the test looks ok?

pyqg/tests/test_parameterizations.py
```diff
@@ -49,5 +49,16 @@ def test_addition_and_scaling(model, rtol=1e-11):
     with pytest.raises(AssertionError):
         back + smag
 
+@pytest.fixture(params=[QG])
+def model(request):
+    klass = request.param()
+    model = klass()
+    model._step_forward()
+    return model
```
asross, line 57: Is there a reason you need to redefine `model` just with `QG`? If it works with other models, then you could just remove this and just include the test you have below.

asross:
Docs look great now, just one small question about the test.

pyqg/tests/test_parameterizations.py
roxyboy: I've never used the parametrization with the other models...

roxyboy: The tests passed locally for my most recent commit for all models.

asross:
Just running CI checks, but everything seems to be in order, so will merge soon!

Might wait a tiny bit to release a new pip version, though, since there are a few other outstanding new parameterization pull requests that we can hopefully get merged first.

PR #318, opened 2023-03-02
Cython updates

karlotness:
This updates some of the flags in the Cython kernel

Cython seems to be in the process of releasing a version 3.0 which deprecates a few things. This should mostly follow their [migration steps](https://cython.readthedocs.io/en/latest/src/userguide/migrating_to_cy30.html)

This also removes the manual chunksize calculation since the default for the static schedule should divide the workload up evenly.

## PyFFTW conditional compilation

One of the things it that seems to be under consideration for deprecation is conditional compilation (with `IF`) used to select pyfftw/np (cython/cython#4310).

The logic for selecting between the two is only run in the `PseudoSpectralKernel.__init__` so we can instead do the selection at runtime.

The existing implementation calls pyFFTW through the Python interface and this doesn't change that.

In the case of the NumPy FFT fallback there might be some slight overhead because of calling through a bound method object, but I would expect that to be small, and users almost certainly use pyFFTW since it is installed by default.

### DIFF
--- a/pyqg/kernel.pyx
+++ b/pyqg/kernel.pyx
@@ -123,7 +123,7 @@ cdef class PseudoSpectralKernel:
         self.ny = ny
         self.nx = nx
         self.nl = ny
-        self.nk = nx/2 + 1
+        self.nk = nx//2 + 1
         self.a = np.zeros((self.nz, self.nz, self.nl, self.nk), DTYPE_com)
         self.kk = np.zeros((self.nk), DTYPE_real)
         self._ik = np.zeros((self.nk), DTYPE_com)
--- a/pyqg/kernel.pyx
+++ b/pyqg/kernel.pyx
@@ -6,9 +6,8 @@
 import numpy as np
 import numpy.fft as npfft
 import warnings
-import cython
 cimport numpy as np
-from cython.parallel import prange, threadid
+from cython.parallel import prange
 
 try:
     import pyfftw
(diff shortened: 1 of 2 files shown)


francispoulin:
@karlotness  I just made a couple of changes to make things consistent with cython, but I see that you have suggested many more.    I am sorry that this is two years late, but I am happy to review your PR.   Is there anything you think we should change it before I review it?

karlotness:
Thanks for taking a look! I've been a bit busy these past few weeks but things are starting to clear up so I can circle back to this a bit.

I think this PR was pretty much ready to go. Most of the updates were just following the Cython migration guide. The larger change was to how PyQG calls into PyFFTW and moving that to a runtime check. That was intended to move away from the conditional compilation feature (the uppercase `IF`) which is deprecated.

I think the changes in the other PRs were intended to be ready to go, but it's possible they were a bit overzealous and could certainly be pared back. I think #317 might still be helpful to clean up some of the test warnings.

The changes in #319 might also help to update some of the setuptools configuration a bit. I think there have been quite a few changes to Python packaging over the years.

However, in that one the changes to OpenMP autodetection might in hindsight not be desirable. I think the thought there was that in combination with #320 PyQG could provide pre-compiled wheels to PyPI. However, those wheels would not be able to use OpenMP even if the system doing the compilation supports it. For that to work (if the pre-compiled builds are wanted) there would need to be some way to force-disable OpenMP during a build (which might be preferable?) or move away from the autodetection.

I have a fork I had made updates to here: https://github.com/karlotness/pyqg/

It includes the changes from the PRs and some other updates, and was just used for some testing. The largest other change was moving the kernel to pure Python, just to make the fork easier to keep working. That comes with a large performance hit though so I wouldn't upstream that one, but I could pick some of the other edits out too.

Thanks for taking a look!

PR #333, opened 2025-05-19
Fix inversion matrix for the BT model

cesar-rocha:
This fixes the inversion matrix `a` for the BT model with free surface.  If a deformation radius is provided, then the inversion matrix `a` is:

` self.a = -1./(self.wv2+self.kd2)`

Otherwise, if  rd is not provided, the model reduces to the 2D vorticity equation, and the `a` is given simply by precomputed inverse wavenumber squared (self.wv2i):

` self.a = -self.wv2i`

### DIFF
--- a/pyqg/bt_model.py
+++ b/pyqg/bt_model.py
@@ -72,8 +72,11 @@ class BTModel(model.Model):
 
     def _initialize_inversion_matrix(self):
         """ the inversion """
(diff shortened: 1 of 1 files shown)


cesar-rocha:
Thanks for reviewing this, @francispoulin. I will merge it.

